PR #168, opened 2026-03-25
Add Claude CLI Server client for Anthropic AI processing

n0an:
## Summary
- Add Claude CLI Server support to route Anthropic AI processing through a Mac running VivaDicta with Claude CLI, using Claude subscription instead of API keys
- New `ClaudeCLIServerClient` HTTP client with enhance, health check, and model fetching endpoints
- Settings UI in Anthropic provider screen: toggle, server URL, auth token, and connection test
- Fallback chain: CLI Server first, then API key if server fails

## Test plan
- [ ] Enable Claude CLI Server toggle in Anthropic settings
- [ ] Enter server URL and auth token from Mac app
- [ ] Tap "Test Connection" — verify green checkmark
- [ ] Run AI processing — verify it routes through CLI Server
- [ ] Disable toggle — verify it falls back to API key
- [ ] Verify local network permission prompt appears on first connection

### DIFF
--- a/VivaDicta/Info.plist
+++ b/VivaDicta/Info.plist
@@ -90,6 +90,8 @@
 		<key>NSAllowsLocalNetworking</key>
 		<true/>
 	</dict>
+	<key>NSLocalNetworkUsageDescription</key>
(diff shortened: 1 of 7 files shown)


n0an:
@claude

PR #177, opened 2026-03-29
Keyboard AI processing: preset picker with VivaMode selector

n0an:
## Summary
- Replace the keyboard's mode-based AI processing with independent VivaMode picker (provider/model) and categorized preset grid
- Add presetId to the IPC layer so the keyboard can request a specific AI preset independently from the VivaMode
- Fix race condition where the main app consumed its own text processing result from UserDefaults before the keyboard extension could read it
- Use liquid glass for category filter chips on iOS 26+
- Fix scroll touch handling in grid gaps using hit-testable backgrounds

## Test plan
- [ ] Open keyboard in any host app, switch to AI processing tab
- [ ] Verify VivaMode picker appears at top and preset grid below with category chips
- [ ] Select a VivaMode, then tap a preset — verify AI processing works
- [ ] Try different presets (built-in and custom) across different VivaModes
- [ ] Verify scrolling works from empty spaces between grid cells
- [ ] Verify category chip filtering works with haptic feedback
- [ ] Verify the feature works with both selected text and text before cursor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -129,3 +129,5 @@ internal/
 
 settings.local.json
 skills-lock.json
+
+.trees/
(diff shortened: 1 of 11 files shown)


n0an:
@claude please review this PR

PR #184, opened 2026-04-04
Add Watch app unit tests, scene phase handling, and testability improvements

n0an:
## Summary

- Add 40 unit tests across 4 test suites (WatchRecordViewModelTests, WatchConnectivityServiceTests, WatchTransferStatusTests, WatchModeInfoTests) with mock implementations
- Inject UserDefaults into WatchRecordViewModel and WatchConnectivityService for test isolation
- Add scene phase handling to stop recording only on `.background` (not `.inactive` which fires during transient overlays)
- Add accessibility identifiers to Watch UI for future UI test automation
- Remove unused WCSessionProtocol (unmockable due to WCSessionFileTransfer having no public init)
- Apply Swift Testing best practices: `#require` for optional unwraps, `== false` instead of `!` in `#expect`
- Add shared Xcode schemes for Watch app and extensions, fix widget scheme to use `accessoryCircular`

## Test plan

- [x] All 40 Watch unit tests pass on watchOS simulator
- [x] Scene phase: recording continues on `.inactive`, stops only on `.background`
- [x] Watch test plan includes both unit and UI test targets
- [x] No regressions in main app tests


n0an:
@claude please review this PR

PR #197, opened 2026-04-13
Restyle keyboard and watch record buttons

n0an:
## Summary
- restyle the keyboard mic button with a dedicated mesh-gradient implementation
- restyle the watch record button with the new mesh ring, glass treatment, and circular glow
- keep watch sizing frame-based so the button stays at half the screen width with a fixed icon size

## Testing
- xcodebuild -scheme VivaDicta -configuration Debug -workspace ./VivaDicta.xcodeproj/project.xcworkspace -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max,OS=26.4' build 2>&1 | xcsift\n\n## Notes\n- the watch view uses watch-local gradient helpers because the shared iPhone gradient view is not part of the watch target\n- no breaking changes

### DIFF
--- a/VivaDicta/Shared/MicButton.swift
+++ b/VivaDicta/Shared/MicButton.swift
@@ -8,36 +8,36 @@
 import SwiftUI
 
 struct MicButton: View {
-    @State var isAnimating = false
-    
+    @State private var isAnimating = false
+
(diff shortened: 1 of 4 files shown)


n0an:
@claude please review this PR

PR #208, opened 2026-04-17
Add Chinese script preference for AI-enhanced output

n0an:
## Summary

- Adds a Settings → AI Settings row letting users choose Simplified or Traditional Chinese for AI-enhanced output
- Preference is appended as a short hint to the AI system message, so it applies across all presets and providers (OpenAI, Anthropic, Gemini, Groq, Apple Foundation Model, etc.)
- Row is locale-conditional: visible only when the device's preferred languages contain Chinese, any mode's transcription language starts with \`zh\`, or the global selected language starts with \`zh\` — no clutter for other users
- When set to the default (\`auto\`), no text is appended — zero behavior change for non-Chinese users

## Motivation

A user reported that Whisper outputs a mix of Simplified and Traditional Chinese characters, and they need consistent Simplified output. Rather than biasing the ASR (Whisper \`promptTokens\` trick is flaky on \`large-v3-turbo\` per WhisperKit #285/#372 and not officially supported), this handles script consistency at the AI stage where LLMs follow instructions reliably. Works equally well with Apple Foundation Model (which distinguishes Simplified/Traditional natively) and cloud LLMs.

## Test plan

- [x] Build passes on iOS 26.4 Simulator
- [x] \`TranscriptionManagerConfigTests\`, \`MistralTranscriptionServiceTests\`, \`PromptsTemplatesTests\` pass (29/29)
- [ ] Manual: device with \`zh-Hans\` preferred language shows the picker; device with only English does not
- [ ] Manual: after creating a mode with Chinese as transcription language on an English device, picker appears
- [ ] Manual: with preference = Simplified, AI enhancement of a Chinese transcription produces Simplified characters (tested across OpenAI + Apple FM)
- [ ] Manual: with preference = Auto, no change to system message vs. main

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -35,7 +35,7 @@ jobs:
         uses: anthropics/claude-code-action@v1
         with:
           claude_code_oauth_token: ${{ secrets.CLAUDE_CODE_OAUTH_TOKEN }}
-          model: claude-opus-4-6
+          claude_args: '--model claude-opus-4-7'
 
           # This is an optional setting that allows Claude to read CI results on PRs
           additional_permissions: |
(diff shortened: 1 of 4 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #210, opened 2026-04-18
Drop preemptive context budgeting for cloud providers

n0an:
## Summary

Stop pretending we know every cloud provider's context window. `ChatContextManager.contextLimit` now returns a real value only for Apple Foundation Models; every cloud provider returns `.max`, which disables preemptive trimming and auto-compaction.

### Why
- The old hardcoded table was lossy and stale: Claude 200K hardcoded while 1M beta exists, OpenAI variants differ, xAI Grok 4 is 256K, etc.
- A too-small entry caused real false positives: Custom OpenAI fell through to the 8K default, so any note above ~5K tokens triggered compaction on open, **even when the configured endpoint was a 200K or 1M model**.
- Preemptive compaction is destructive - we summarize history/context that didn't need summarizing. Being wrong here is worse than doing nothing.
- Cloud providers already return clean `context_length_exceeded` errors. Users can compact manually via the existing **Compact** button if a conversation grows too long.

### What changes
- `ChatContextManager.contextLimit` collapses to Apple + default `.max`.
- Preemptive `shouldAutoCompact` calls removed from `ChatViewModel.sendCloudMessage`, `MultiNoteChatViewModel.sendCloudMessage`, `SmartSearchChatViewModel.sendCloudMessage`.
- `shouldAutoCompact` helpers deleted from `ChatContextManager`, `MultiNoteContextManager`, `SmartSearchContextManager` (no more callers).
- Cloud branch of `updateContextFillRatio` removed in all three view models - cloud just sets ratio to 0.
- `%` pill in the three chat views is hidden when ratio is 0, so cloud conversations no longer show a misleading `0%`.

### What does not change
- Apple FM still uses real `SystemLanguageModel.default.contextSize`, still shows the fill pill, still auto-compacts at 70%, still has reactive compaction on `exceededContextWindowSize`.
- `assembleMessages` still exists; with `limit = .max` it simply packs everything (no note truncation, no message drop).
- The manual **Compact** button works as before for cloud.

### Net
9 files, -139 / +49 lines.

## Test plan
- [x] Clean build passes on iPhone 17 Pro Max simulator, iOS 26.4 (0 errors, 29 pre-existing warnings).
- [ ] Smoke: open a long single-note chat on a cloud provider; confirm no "auto-compacting" log line fires at open and the % pill is hidden.
- [ ] Smoke: same on multi-note chat and Smart Search chat.
- [ ] Smoke: Apple FM path still shows the % pill and still auto-compacts past 70%.
- [ ] Smoke: Custom OpenAI against a local endpoint - open a note that previously triggered preemptive compaction; confirm it no longer does.

### DIFF
--- a/VivaDicta/Services/AIEnhance/ChatContextManager.swift
+++ b/VivaDicta/Services/AIEnhance/ChatContextManager.swift
@@ -69,50 +69,23 @@ struct ChatContextManager {
 
     // MARK: - Context Limits
 
-    /// Known context window sizes per provider/model prefix.
-    /// Returns token limit for the given provider and model.
+    /// Context window in tokens for the given provider and model.
+    ///
+    /// Only Apple Foundation Models has a real, knowable, and actionable limit:
+    /// the model runs on-device and we must pre-flight budget before sending.
+    /// For every cloud provider the window varies by model and evolves faster
+    /// than our hardcoded table can track, and the server already enforces its
+    /// own limit and returns a clean error if we overshoot. Returning `.max`
+    /// disables preemptive trimming and auto-compaction for cloud.
     static func contextLimit(for provider: AIProvider, model: String) -> Int {
-        let modelLower = model.lowercased()
-
         switch provider {
         case .apple:
             if #available(iOS 26, *) {
                 return SystemLanguageModel.default.contextSize
             }
             return 4_096
-
-        case .anthropic:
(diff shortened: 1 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #211, opened 2026-04-18
Expose transcribed notes as Shortcuts outputs

n0an:
## Summary
- Add `GetLatestTranscriptionIntent` - returns the most recent note as a chainable `TranscriptionEntity`, so users can pipe its `Text` or `Enhanced Text` into other Shortcuts actions.
- Add `RecordAndReturnTranscriptionIntent` - a single-step intent that starts recording via `AppGroupCoordinator`, waits for transcription to finish (polls SwiftData for a new `Transcription` with `timestamp > start`), and returns the resulting entity.
- Register both in `ShortcutsProvider` with Siri phrases (all include `\(.applicationName)`).

Polling reads SwiftData rather than installing a second listener on `AppGroupCoordinator.onTranscriptionCompleted`, because that callback is single-owner and already wired to `RecordViewModel`. `RecordViewModel` saves the `Transcription` before calling `shareTranscribedText`, so the DB is the correct source of truth.

Motivation: users want to use VivaDicta's transcription inside their own Shortcuts flows (e.g. append to Obsidian via `obsidian://new?file=<date>&clipboard&append=true`) without relying on a third-party OpenAI call.

## Test plan
- [x] Clean build succeeds (`xcodebuild` on iPhone 17 Pro Max sim, iOS 26.4, 0 errors).
- [ ] In Shortcuts, run `Get Latest Note` and verify it returns a note with `Text` / `Enhanced Text` fields.
- [ ] In Shortcuts, run `Record and Get Note`, dictate, tap stop, verify the returned entity matches the newly created transcription.
- [ ] Repro the friend's Obsidian flow: Record and Get Note -> Copy `Enhanced Text` to Clipboard -> Open `obsidian://new?file=<date>&clipboard&append=true`.
- [ ] Error path: start an intent, cancel the recording, verify it times out at 180s with a localized dialog instead of hanging.

PR #212, opened 2026-04-18
Debounce record-and-return cancel detection

n0an:
## Summary
Fixes Option A of the Obsidian/Shortcuts flow introduced in #211 falsely throwing "Recording was cancelled." when the user taps Stop inside VivaDicta.

Root cause: `RecordViewModel` synchronously calls `AppGroupCoordinator.updateRecordingState(false)` on Stop, but only later (inside an async `Task { @MainActor }`) writes `transcriptionStatus = .transcribing`. The intent's 500ms poll loop catches that one-to-two-tick window where `!isRecording && status == .idle` and trips the cancel detector.

Fix: debounce - require six consecutive idle ticks (~3s) before throwing `.cancelled`. Real cancels persist indefinitely; the stop-transition gap is at most one or two ticks.

## Test plan
- [x] Build succeeds (xcodebuild iPhone 17 Pro Max sim iOS 26.4, 0 errors).
- [ ] Option A shortcut: Record and Get Note -> Copy Enhanced Text -> Open Obsidian URL. Dictate 5s, tap Stop in VivaDicta, confirm the shortcut proceeds with the new text.
- [ ] Option B shortcut still works (Start -> Alert -> Stop -> Wait -> Get Latest).
- [ ] User cancels recording in-app: shortcut still throws Recording was cancelled within ~3s instead of hanging 180s.

### DIFF
--- a/VivaDicta/AppIntents/RecordAndReturnTranscriptionIntent.swift
+++ b/VivaDicta/AppIntents/RecordAndReturnTranscriptionIntent.swift
@@ -72,6 +72,7 @@ struct RecordAndReturnTranscriptionIntent: AppIntent {
     ) async throws -> TranscriptionEntity {
         let deadline = Date().addingTimeInterval(Self.maxWaitSeconds)
         var sawActiveSession = false
+        var consecutiveIdleTicks = 0
 
         while Date() < deadline {
             try Task.checkCancellation()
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #222, opened 2026-04-20
Keep GitHub Copilot OAuth poll alive when app backgrounds

n0an:
## Summary
- Wrap the device-code poll loop in a `UIApplication` background task so polling continues ~30s while Safari is foregrounded for the GitHub authorization prompt.
- Catch transient `URLError` (e.g. `-1005 networkConnectionLost`) inside the poll loop and retry on the next interval instead of aborting. `URLError.cancelled` still propagates so the Cancel button works.

## Why
Users were seeing "Sign-In Error - The network connection was lost" after returning from Safari because iOS tore down the `URLSession.shared` connection while the app was backgrounded. One transient failure killed the whole flow. With these two fixes, the typical 15-60s Safari round-trip finishes with no user-visible error.

## Test plan
- [x] Debug build succeeds
- [ ] Sign into GitHub Copilot, enter device code in Safari, return to app -> "Connected" appears, no error
- [ ] Cancel button during polling stops the flow cleanly
- [ ] If Safari takes >30s, next foreground tick retries and completes successfully

### DIFF
--- a/VivaDicta/Services/OAuth/CopilotOAuthManager.swift
+++ b/VivaDicta/Services/OAuth/CopilotOAuthManager.swift
@@ -73,6 +73,14 @@ final class CopilotOAuthManager: Sendable {
         let pollInterval = max(interval, 5)
         let maxAttempts = expiresIn / pollInterval
 
+        // Keep polling alive for ~30s while Safari is foregrounded for the device-code prompt.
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #223, opened 2026-04-21
Support chat for OpenAI OAuth via Codex Responses API

n0an:
## Summary

- OAuth-only OpenAI users can now chat in single-note, multi-note, and Smart Search surfaces without adding an API key
- Chat routes through the Codex Responses backend (`chatgpt.com/backend-api/codex/responses`) that the app already uses for transcription enhancement
- Request shape is the structured Codex form: system → `instructions`, turns → `input` items with `input_text`/`output_text` content blocks, `store: false`, `stream: true`
- Planner-based tools (Search other notes, Search web) work via `makeChatRequest(...)` automatically; implicit mid-chat tool-calling stays deferred

## Changes

- `OpenAIOAuthClient` gains `chatStreaming(...)` + `chat(...)` and extracts shared `buildResponsesRequest(...)` / `streamResponsesText(...)` helpers. `enhance(...)` behavior unchanged in terms of output.
- `AIService+Chat.swift` adds `.openAIOAuth` to `ChatStreamingRoute` and branches in both `makeChatStreamingRequest` and `makeChatRequest`. Removes the "Chat is not yet supported with OpenAI OAuth" error. Gemini OAuth gate kept.
- Auto-fallback to API key on `OAuthError` when both OAuth and API key are configured.
- Chat output runs through `AIEnhancementOutputFilter.filter(...)` to match the API-key path (strips `<thinking>` / XML wrappers if emitted).
- HTTP errors from the Codex backend are now classified: 401/403 → `OAuthError` (fallback fires), 429 → `EnhancementError.rateLimitExceeded`, 5xx → `EnhancementError.serverError`, other → `EnhancementError.customError`. Prevents silent API-key quota burn on transient server blips.
- `documentation/RAG/Tool-Provider-Support.md` updated to reflect the new OAuth chat capability and the deferred implicit tool-calling.

## Behavior change to note

Previously, a user with both OAuth *and* an API key got the API-key chat path. After this PR, they get OAuth chat. This matches how enhancement already behaves - if the user explicitly signed in with OAuth, that intent is respected. API key remains as an automatic fallback when OAuth errors out.

## Deferred

- Implicit mid-chat tool-calling for OAuth (`makeCrossNoteSearchToolDecision`, `makeWebSearchToolDecision`). The existing catch blocks in chat view models silently skip implicit tools on error, so OAuth users just don't get them - no user-visible breakage.
- Gemini OAuth chat - same structural pattern, will be a separate PR.
- Rare mid-stream token expiry edge case: if a token dies partway through a streaming response, the current fallback restarts the request and the UI would see the stream restart. Very rare (tokens valid for ~1h; we refresh proactively before dispatch).
- Unit tests for the new route - separate PR.

## Test plan

- [x] Build passes on iOS Simulator (iPhone 17 Pro Max, iOS 26.4)
- [x] Single-note chat with OpenAI OAuth
- [x] Cross-note search from chat with OpenAI OAuth
- [ ] Multi-note chat with OpenAI OAuth
- [ ] Smart Search chat with OpenAI OAuth
- [ ] Web search from chat with OpenAI OAuth
- [ ] Cancellation mid-stream
- [ ] Fallback to API key when OAuth token is invalid
- [ ] No regression for OpenAI API-key-only accounts
- [ ] No regression for single-turn transcription enhancement over OpenAI OAuth

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService+Chat.swift
+++ b/VivaDicta/Services/AIEnhance/AIService+Chat.swift
@@ -129,11 +129,34 @@ extension AIService {
                 onPartialResponse: onPartialResponse
             )
 
-        case nil:
-            // Check if this is an OAuth-only provider without chat support
-            if provider == .openAI && isOpenAISignedIn && provider.apiKey == nil {
-                throw EnhancementError.customError("Chat is not yet supported with OpenAI OAuth. Please add an OpenAI API key to use chat.")
+        case .openAIOAuth:
+            do {
+                let oauthProvider = OpenAIOAuthProvider()
+                let (token, accountId, _) = try await OAuthManager.shared.validAccessToken(for: oauthProvider)
+                let resolvedModel = OpenAIOAuthClient.resolveModel(model)
+                return try await OpenAIOAuthClient.chatStreaming(
+                    systemMessage: systemMessage,
+                    messages: messages,
+                    model: resolvedModel,
+                    accessToken: token,
+                    accountId: accountId,
+                    onPartialResponse: onPartialResponse
+                )
+            } catch let error as OAuthError {
+                if let apiKey = provider.apiKey {
+                    return try await makeOpenAIChatStreamingRequest(
+                        url: URL(string: provider.baseURL)!,
+                        model: model,
+                        systemMessage: systemMessage,
+                        messages: messages,
+                        headers: ["Authorization": "Bearer \(apiKey)"],
+                        onPartialResponse: onPartialResponse
+                    )
+                }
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Fixed in 95c530d..HEAD - `chatStreaming` now runs `AIEnhancementOutputFilter.filter` on the final result and re-emits through the partial callback so the UI updates if filtering stripped anything. Matches what the API-key path already does.

PR #225, opened 2026-04-21
Add AZERTY keyboard layout option

n0an:
## Summary
- Adds a Layout picker (QWERTY / AZERTY) in Settings > Keyboard, persisted via `AppGroupCoordinator` and shared with the keyboard extension.
- Rewrites the three iPhone letter rows of KeyboardKit's standard layout to AZERTY (`a z e r t y u i o p` / `q s d f g h j k l m` / `w x c v b n`) while preserving shift, backspace, space, return and per-key sizes.
- Supplies French long-press callouts (`é è ê ë`, `à â æ`, `ç`, `ñ`, etc.) via `.keyboardCalloutActions`, since KeyboardKit Pro's locale-specific input sets and callout tables are license-gated.
- Observes `KeyboardContext` from the keyboard view so shift cycles correctly: `KeyboardAction.shift` carries the current case in its associated value and the standard layout swaps `.character("a")` / `.character("A")` per render, so a frozen custom layout would break shift. For QWERTY we now pass `layout: nil` so KeyboardKit keeps full control; for AZERTY we rebuild each pass and mirror the standard layout's case onto the replaced chars.

## Scope / Limitations
- iPhone only. On iPad the shape differs (4 letter rows, extra keys) and the transformer falls back to QWERTY silently rather than produce a broken row.
- No locale auto-switch: user picks layout manually.

## Test plan
- [ ] Select AZERTY in Settings; open the keyboard - letter rows show A Z E R T Y U I O P / Q S D F G H J K L M / W X C V B N.
- [ ] Tap shift in AZERTY - letters switch to uppercase and a second tap returns to lowercase; caps-lock (double-tap) behaves as expected.
- [ ] Long-press `e` in AZERTY - callout shows `é è ê ë ę`; selecting one inserts it. Same for `a`, `c`, `i`, `n`, `o`, `u`, `y`.
- [ ] With shift active, long-press `E` - callout shows `É È Ê Ë Ę`.
- [ ] Switch back to QWERTY - layout returns to standard, shift cycles correctly, English callouts apply.
- [ ] Dictation flow (mic button, text processing, recent notes tabs) unaffected by layout change.


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Fixed the P1 inline on AzertyLayout.swift:45 in 40de85838 - the rewrite is now gated on `keyboardContext.keyboardType == .alphabetic`, so the numeric and symbolic keyboards are left alone. Thanks @codex.

n0an:
@claude please review this PR again - your previous attempt errored out.

PR #226, opened 2026-04-24
Add per-mode Obsidian append integration

n0an:
## Summary

Adds an opt-in per-mode setting to append each transcription to an Obsidian note via the `obsidian://new?file=...&clipboard&append=true` URL scheme. Works from both the main app (post-transcription hook in `RecordViewModel`) and the keyboard extension (verified that SwiftUI's `@Environment(\.openURL)` can open external schemes from a keyboard extension, same mechanism used by the existing `vivadicta://` deep link).

- New \"Obsidian\" section in the mode editor with a toggle + note name template (defaults to one note per transcription: \`VD {date} {HH}-{mm}-{ss}\`).
- Shared \`ObsidianURLBuilder\` helper consumed by both the main app and the keyboard target (added to keyboard membership exceptions in pbxproj).
- Placeholders: \`{date}\`, \`{yyyy}\`, \`{MM}\`, \`{dd}\`, \`{HH}\`, \`{mm}\`, \`{ss}\`, \`{preset}\`, \`{mode}\`. Set the template to just \`{date}\` to get daily-note append behaviour instead.
- Date placeholders are forced to Gregorian so \`{date}\` always resolves to \`YYYY-MM-DD\` regardless of the user's iOS calendar setting; formatted with \`IntegerFormatStyle\` (no \`String(format:)\`).
- Backward-compatible decoding on \`VivaMode\` - existing saved modes load with the toggle off.

Feature is opt-in. There is a temporary purple flask probe button in the keyboard toolbar that was used to validate the extension URL-open path; it can be removed once end-to-end device testing confirms the real hook works.

## Test plan

- [ ] Build & run on a physical device (simulator cannot hand off to Obsidian).
- [ ] Install Obsidian on the device and open any vault.
- [ ] Enable \`Append to Obsidian\` on a mode with defaults.
- [ ] Main app: record + enhance -> VivaDicta backgrounds and a new note appears in Obsidian.
- [ ] Keyboard: swap to VivaDicta keyboard in a text field, tap mic -> text inserts AND Obsidian foregrounds with a new note.
- [ ] Flip template to \`{date}\`, record twice -> both entries land in the same daily note.
- [ ] Turn the toggle off -> footer hides, no Obsidian hand-off happens.
- [ ] Open an existing pre-upgrade mode -> Obsidian section renders with toggle off, no behaviour change.
- [ ] Remove the temp flask probe button once happy with behaviour.

### DIFF
--- a/VivaDicta.xcodeproj/project.pbxproj
+++ b/VivaDicta.xcodeproj/project.pbxproj
@@ -335,6 +335,7 @@
 				Shared/AppGroupCoordinator.swift,
 				Shared/KeyboardLayoutStyle.swift,
 				Shared/MicButton.swift,
+				Shared/ObsidianURLBuilder.swift,
 				Shared/RecentNotesCache.swift,
 				Shared/RecordingState.swift,
 				Shared/RippleEffect.metal,
--- a/VivaDicta/Services/AIEnhance/VivaMode.swift
+++ b/VivaDicta/Services/AIEnhance/VivaMode.swift
@@ -82,14 +82,8 @@ struct VivaMode: Identifiable, Hashable, Codable {
     /// Whether to append the transcription to an Obsidian note after completion.
     var obsidianEnabled: Bool
 
-    /// Optional Obsidian vault name. Empty = use last-opened vault.
-    var obsidianVault: String
-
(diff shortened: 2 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #227, opened 2026-04-25
Add Settings → Integrations with global Obsidian toggle

n0an:
## Summary

Promotes the Obsidian feature out of the per-mode editor and into a dedicated **Settings → Integrations** screen, so it can scale to other integrations (Webhooks, Zapier, etc.) without cluttering the mode form.

- New \`IntegrationsView\` (Settings → Integrations) hosts the master Obsidian toggle, the note-name template, and the explanatory footer. Reserved as the home for future integration rows.
- Two new app-private \`UserDefaults\` keys: \`isObsidianGloballyEnabled\` (default off) and \`obsidianNoteTemplate\` (default \`VD {date} {HH}-{mm}-{ss}\`). Keyboard never reads these - it just consumes the App Group hand-off the main app stashes.
- \`VivaMode.obsidianNoteTemplate\` is removed (template is now global). \`VivaMode.obsidianEnabled\` stays as a per-mode opt-out, default flipped to \`true\` so flipping the global toggle on auto-enables every mode.
- \`ModeEditView\`'s Obsidian section is now a single toggle that only renders when the global integration is on.
- \`RecordViewModel.openObsidianIfEnabled\` short-circuits when the global flag is off and reads the template from the global setting.
- \`ObsidianURLBuilder.build\` now takes \`template:\` + \`modeName:\` instead of the whole \`VivaMode\`.

The integration shipped to main yesterday so there's no installed-base mode JSON with these fields - simple decoder defaults are sufficient (no migration logic).

## Test plan

- [ ] Fresh install: Settings → Integrations shows Obsidian toggle off, no template field. Mode editor has no Obsidian section.
- [ ] Flip global toggle on: template field appears with default \`VD {date} {HH}-{mm}-{ss}\` placeholder. Mode editor now shows a per-mode \"Save to Obsidian\" toggle, on by default.
- [ ] Record from main app → Obsidian opens directly with the correct content.
- [ ] Record from keyboard (Hot Mic) → main app delegates via App Group, keyboard opens Obsidian with the correct content.
- [ ] Edit the global template (e.g. \`{date}\`) → next transcription uses the new template.
- [ ] Disable per-mode toggle on one mode, leave others on → only that mode skips Obsidian.
- [ ] Flip global off → mode editor hides the per-mode toggle; subsequent transcriptions skip Obsidian regardless of per-mode state.

### DIFF
--- a/VivaDicta/Services/AIEnhance/VivaMode.swift
+++ b/VivaDicta/Services/AIEnhance/VivaMode.swift
@@ -79,11 +79,10 @@ struct VivaMode: Identifiable, Hashable, Codable {
     /// Whether smart insert (auto-adjust spacing and capitalization) is applied when inserting text via keyboard.
     let isSmartInsertEnabled: Bool
 
-    /// Whether to append the transcription to an Obsidian note after completion.
+    /// Whether this mode opts in to saving transcriptions to Obsidian.
+    /// Only effective when the global Obsidian integration is enabled in
+    /// Settings → Integrations.
     var obsidianEnabled: Bool
-
-    /// Template for the Obsidian note filename. Placeholders: {date}, {yyyy}, {MM}, {dd}, {HH}, {mm}, {ss}, {preset}, {mode}.
-    var obsidianNoteTemplate: String
 
     /// Creates a new VivaMode with the specified settings.
     init(id: UUID,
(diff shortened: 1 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #231, opened 2026-04-29
Fix Stop button hang in Live Translation

n0an:
## Summary
The Live Translation Stop button could freeze the UI for several seconds or indefinitely. Both `LiveTranslationService` and `LiveTranslationAudio` are `@MainActor`, so when `audio.stop()` ran the AVAudioEngine + AVAudioSession teardown synchronously, MainActor was blocked for the duration. SwiftUI couldn't process state updates or tap events, so the user perceived it as "app hangs forever, only restart helps".

## Changes

- `audio.stop()` is now `async`. The hardware teardown (`playerNode.stop`, `inputNode.removeTap`, `engine.stop`, `AVAudioSession.setActive(false)`) runs in a `Task.detached` wrapped in a small `@unchecked Sendable` `AudioTeardownContext` class. MainActor is yielded for the duration; SwiftUI continues to process updates and taps while the engine cleans up. AVAudioEngine and AVAudioSession are documented as thread-safe for these operations - the wrapper just satisfies Swift 6 strict concurrency.
- Drop `.notifyOthersOnDeactivation` from `setActive(false)`. It triggers synchronous notifications to other apps that have been observed to take seconds in some iOS scenarios; we don't strictly need it for this feature.
- Per-step diagnostic logging in `LiveTranslationService.teardown()` and `AudioTeardownContext.perform()`. If the hang ever resurfaces, the device console (subsystem `com.antonnovoselov.VivaDicta`, categories `LiveTranslationService` / `LiveTranslationAudio`) will print exactly which step is stalled so we can target that specific call.

## Test plan
- [ ] Open Live Translation, tap Start.
- [ ] Speak / play a short audio clip so STT and TTS produce real activity.
- [ ] Tap Stop. UI should respond within ~1 second; the Save button should appear.
- [ ] Repeat 3-4 Start -> Stop cycles in the same screen session.
- [ ] Long-running session (~5 minutes), Stop should still respond promptly.
- [ ] No headphones case (held to ear / earpiece). Stop should still work.
- [ ] If hang resurfaces, capture Console.app filtered to `LiveTranslationAudio` to identify which teardown step stalled.

## Notes
This was already pushed to `main` as `558cc9065` and then reverted out of `main` so it could go through normal PR review. The branch carries the same commit; merging this PR re-introduces the fix.

### DIFF
--- a/VivaDicta/Services/LiveTranslation/LiveTranslationAudio.swift
+++ b/VivaDicta/Services/LiveTranslation/LiveTranslationAudio.swift
@@ -97,7 +97,7 @@ final class LiveTranslationAudio {
             playerNode.play()
             isStarted = true
         } catch {
-            stop()
+            await stop()
             throw error
         }
     }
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #236, opened 2026-04-29
Add Speechmatics as a cloud transcription provider with inline translation

n0an:
## Summary

Adds Speechmatics' batch API as a cloud transcription provider. Speechmatics handles transcription, speaker diarization, and target-language translation in a single async job, so the existing per-mode "Translate to" picker now lights up for Speechmatics alongside Soniox and Gladia.

### Service
- New `SpeechmaticsTranscriptionService` implements Speechmatics' 3-step batch flow:
  1. `POST /v2/jobs` (multipart with `config` JSON + `data_file` audio) → job `id`
  2. Poll `GET /v2/jobs/{id}` at 2Hz until `job.status == "done"` (or `"rejected"`); 5-minute hard timeout
  3. `GET /v2/jobs/{id}/transcript?format=json-v2` → parse `results[]` (one item per word) + optional `translations`
- Plain text is rebuilt from per-word `alternatives[0].content`, attaching punctuation tokens to the previous word without an extra space
- Diarization walks `alternatives[].speaker` per word and merges consecutive same-speaker words into turns
- Translation reads from top-level `translations.<target>[]` sentence segments. If a target is requested but missing or empty, the service throws `noTranscriptionReturned` instead of silently returning the original transcript (matching the Gladia pattern from the previous PR)
- Custom vocabulary passes through as `additional_vocab: [{content: term}]`
- `zh` → `cmn` mapping at request time, since Speechmatics uses `cmn` for Mandarin while we (and most other providers) use `zh`. Users still see "Chinese" in the picker

### Provider plumbing
- `TranscriptionModelProvider.speechmatics` and `AIProvider.speechmatics` enum cases with display name, icon (`speechmatics` asset), default model `speechmatics-batch-v2`, base URL `https://asr.api.speechmatics.com/v2`, console URL `https://portal.speechmatics.com/manage-access/`, keychain key `speechmaticsAPIKey`
- New `CloudModel "speechmatics-batch-v2"` backed by `speechmaticsLanguages` (51 standard codes; bilingual codes like `ar_en`, `cmn_en` and the multilingual `cmn_en_ms_ta` are intentionally out of scope - they don't fit the existing language picker)
- `CloudTranscriptionService` routes `case .speechmatics`
- `AIService.verifySpeechmaticsAPIKey` does a cheap `GET /v2/jobs?limit=1` with `Authorization: Bearer`
- `APIKeyMigrationService` migrates the keychain entry
- `LoggerExtension` adds `speechmaticsTranscriptionService` category

### UI
- `ModeEditViewModel.isTranslationTargetSelectionAvailable()` now also returns true for `.speechmatics`
- Speaker Labels help text in Settings now lists Speechmatics
- `VivaMode.translationTargetLanguage` doc comment updated

### Out of scope (deliberately)
- **Native summarization** - Speechmatics supports built-in summarization (`summarization_config: { content_type, summary_length, summary_type }`). VivaDicta already has equivalent AI Processing presets ("Summary", "Action Points") that work across all providers via OpenAI/Anthropic/etc., so duplicating that surface for one provider isn't worth the data-model changes right now. If we revisit this, it should be a generic mechanism shared with Gladia (which also has native summarization), not Speechmatics-only.
- **Bilingual / multilingual language codes** (`ar_en`, `cmn_en`, `cmn_en_ms_ta`)
- **Real-time WebSocket transcription** - separate scope from batch

## Test plan

- [x] Build succeeds (`xcodebuild ... build`)
- [ ] Add a Speechmatics API key in Settings → AI Providers → key is verified and saved
- [ ] Create a mode using Speechmatics + `speechmatics-batch-v2` → transcribe a file, verify text comes back with proper punctuation spacing
- [ ] Same mode + speaker diarization on → verify per-speaker output
- [ ] Same mode + "Translate to: French" on a non-French recording → verify only the French translation is returned
- [ ] Speechmatics + diarization + translation combined → verify diarized translated output
- [ ] Pick "Chinese" as source language → verify the request actually sends `cmn` (not `zh`)
- [ ] Pick a language Speechmatics doesn't support (e.g. Armenian) → confirm it isn't in the picker


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #238, opened 2026-05-01
Add per-mode Trim Trailing Period toggle

n0an:
## Summary

- Adds a per-mode **Trim Trailing Period** toggle in Edit Mode → Text Processing (right below Smart Insert) that strips trailing \`.\` or \`...\` from the final transcript.
- Mirrors Python's \`text.strip().rstrip('.')\` - eats any number of trailing periods, leaves \`!\` and \`?\` alone, and never touches internal periods (e.g. acronyms like \`J.R.R. Tolkien\` pass through unchanged).
- Default **off**; existing users see no behavior change.
- Runs as the last step of the post-transcription pipeline in \`TranscriptionManager\` so it operates on the final user-visible text after replacements.

## Why

Customer request from a VivaDicta user (Peter Szemraj) who dictates casual messages 90%+ of the time. In chat contexts, terminal periods read as cold or passive-aggressive (well-documented in psycholinguistics - "Okay." reads colder than "Okay" because the message bubble already marks the end of the utterance). Per-mode means he can have it on for the messaging mode and off for notes/email modes.

## Smart Insert interaction

Traced the keyboard-side flow. \`TextInsertionFormatter\` only modifies whitespace and capitalization - it never inserts or removes periods. The two features compose cleanly:

- **Strip on + Smart Insert on:** consistent casual style. Stripping the period also causes Smart Insert to lowercase the next dictation (since it caps after \`.!?\\n\`), so multi-message flows stay in the casual register.
- **Strip off + Smart Insert on:** existing formal style preserved.

## Test plan

- [x] 9 new unit tests in \`TranscriptionOutputFilterTests\` covering single-sentence strip, multi-sentence (only trailing period removed), ellipsis (stripped per spec), \`?\` and \`!\` preserved, no-op on missing trailing period, empty string, trailing-whitespace handling, internal periods untouched (acronym test). All 33 tests pass.
- [x] Clean \`xcodebuild\` for iPhone 17 Pro Max, OS 26.4 simulator.
- [x] Codable round-trip is backwards-compatible: \`decodeIfPresent ?? false\` for the new field, so existing serialized modes load without migration.

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -301,7 +301,8 @@ class AIService {
             useClipboardContext: mode.useClipboardContext,
 
             isAutoTextFormattingEnabled: mode.isAutoTextFormattingEnabled,
-            isSmartInsertEnabled: mode.isSmartInsertEnabled
+            isSmartInsertEnabled: mode.isSmartInsertEnabled,
+            isStripTrailingPeriodEnabled: mode.isStripTrailingPeriodEnabled
         )
 
         addMode(duplicatedMode)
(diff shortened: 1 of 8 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Fixed in 221071e36 - thanks @codex, real catch. All six `VivaMode(...)` reconstruction sites in `AIService` now explicitly pass `mode.isStripTrailingPeriodEnabled`. Build is green.

Side note: the same code shape also drops `obsidianEnabled` (none of the reconstructions pass it, so it silently resets to its default `true`). That's a pre-existing bug unrelated to this PR - flagging for a follow-up.

PR #242, opened 2026-05-07
Add ASCII recording orb as alternative visualization

n0an:
## Summary

- New `ASCIIOrbView` component (40x40 monospaced grid, sphere SDF + head-on diffuse + radial sine waves + 2-octave value noise) renders an audio-reactive orb as ASCII characters. Same `audioPower` binding as `ParticleOrbView` so it's a drop-in.
- **Defaults to enabled** for both new and existing users. Existing users who never opted out will see the ASCII orb on first launch after this ships; they can switch back via Settings -> Feedback -> ASCII Recording Orb.
- `RecordingSheetView` branches on the toggle and renders either `ASCIIOrbView` or the original `ParticleOrbView`.
- Visual refinements: head-on lighting so the brightest cell is at the orb's geometric center; explicit radial sine waves at two frequencies so ripples genuinely propagate from the center outward; soft, noise-perturbed silhouette (smoothstep alpha through `effRNorm = 0.80...1.15`) so the edge feels fluid instead of stamped at a perfect circle.
- Color: cool->warm gradient via inline RGB lerp (avoids `Color.resolve(in:)` overhead at ~1500 cells per frame).
- Performance: `Canvas(rendersAsynchronously: true)` + per-character `resolve` once per frame with mutated `shading` per cell; cells outside the silhouette band and below a lightness threshold are skipped.

## Test plan

- [ ] Build succeeds on iPhone 17 Pro Max simulator (iOS 26.4)
- [ ] Fresh install / first launch shows ASCII orb on the recording sheet
- [ ] Toggle off in Settings -> recording sheet falls back to the original `ParticleOrbView`
- [ ] Orb pulses/intensity tracks audio power
- [ ] Ripples visibly emerge from the orb's center and roll outward
- [ ] Edge silhouette wobbles softly rather than clipping at a hard circle
- [ ] Toggle survives app relaunch (UserDefaults persists)
- [ ] Frame rate feels smooth on physical device while recording

### DIFF
--- a/VivaDicta/Shared/UserDefaultsStorage.swift
+++ b/VivaDicta/Shared/UserDefaultsStorage.swift
@@ -84,6 +84,9 @@ enum UserDefaultsStorage {
         // What's New
         static let lastSeenWhatsNewVersion = "lastSeenWhatsNewVersion"
 
+        // Recording sheet visualization
+        static let isASCIIOrbEnabled = "isASCIIOrbEnabled"
+
         // Notes filter
         static let savedNotesFilterSourceTags = "savedNotesFilterSourceTags"
         static let savedNotesFilterUserTagIds = "savedNotesFilterUserTagIds"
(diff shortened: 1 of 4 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #244, opened 2026-05-08
Temporary AI mode override in AI Actions sheet

n0an:
## Summary
- Adds a mode chip in the top-left of the AI Actions sheet so the user can swap the AI brain (provider/model) for a single preset tap without touching the active mode. Override is local to the sheet and resets on dismiss.
- When the active mode has no AI configured but other AI-enabled modes exist, the sheet opens directly into the existing "AI Processing Not Configured" empty state with the chip available. Picking an override mode swaps the body to the preset list and hides the "Open Mode Settings" button.
- The legacy `ConfigureAISheet` only appears in the cold-start case where no mode anywhere has AI configured.

## Implementation notes
- `AIService.makeRequest` / `makeStreamingRequest` and request-path helpers (`makeOllamaRequest`, `makeAnthropicStreamingRequest`, `resolvedStreamingRoute`) now thread an explicit `mode: VivaMode` parameter instead of reading `selectedMode`. The observable `selectedMode` is never mutated for an override, so SwiftUI observers do not flicker mid-request.
- `generateVariation(text:preset:modeOverride:onPartialResult:)` accepts an optional override and falls back to `selectedMode` when nil. `enhance(_:)` continues to pass `selectedMode` through.
- New public `supportsResponseStreaming(for: VivaMode) -> Bool`; `currentModeSupportsResponseStreaming` becomes a wrapper.
- `TranscriptionVariation` records the override mode's `aiModel` and `aiProvider` so history attribution is correct when comparing outputs.

## Test plan
- [ ] Build clean on iPhone 17 Pro Max simulator (verified locally).
- [ ] Active mode has AI: open AI Actions sheet, confirm chip shows "Default" gray; pick a different mode, confirm chip turns accent-tinted.
- [ ] Tap a preset with override active; confirm the resulting variation chip displays the override mode's model/provider in metadata.
- [ ] Active mode lacks AI but another mode has AI: tap AI button, confirm "Not Configured" copy + Open Mode Settings + chip in corner. Pick override -> body morphs to preset list, button disappears. Pick "Default" again -> reverts to empty state.
- [ ] No mode has AI anywhere: tap AI button, confirm legacy `ConfigureAISheet` (no chip) still appears.
- [ ] Streaming providers still stream when override picks a streaming-capable model; non-streaming fallback works otherwise.
- [ ] Regenerate-from-chip path (no override) still uses `selectedMode`.

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -633,22 +633,24 @@ class AIService {
     /// - A prompt with instructions is selected
     ///
     /// - Returns: `true` if enhancement can proceed, `false` otherwise.
-    public func isProperlyConfigured() -> Bool {
+    public func isProperlyConfigured(for modeOverride: VivaMode? = nil, requirePreset: Bool = true) -> Bool {
+        let mode = modeOverride ?? selectedMode
+
         // Check if AI processing is enabled
-        guard selectedMode.aiEnhanceEnabled else {
-            logger.logInfo("AI processing is disabled for mode: \(self.selectedMode.name)")
+        guard mode.aiEnhanceEnabled else {
+            logger.logInfo("AI processing is disabled for mode: \(mode.name)")
             return false
         }
 
         // Check if AI provider is selected
-        guard let aiProvider = selectedMode.aiProvider else {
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #246, opened 2026-05-09
Add manual Send to Obsidian button on transcription detail

n0an:
## Summary

Splits the single "Save to Obsidian" toggle in **Settings → Integrations** into two independent siblings, addressing user feedback that the auto-open behavior is intrusive.

- **Auto-open after transcription** - existing behavior, storage key `isObsidianGloballyEnabled` retained so users who had it on stay on.
- **Show Send to Obsidian button** - new key `isObsidianSendButtonEnabled`. When enabled, a circular FAB appears on the transcription detail screen (to the left of the existing Append-with-Voice mic button) and sends the currently displayed text - either the original or the selected variation - to Obsidian on demand.

The "Note name" template field + footer copy now show when **either** toggle is on. The manual send shares the same template; the per-mode `obsidianEnabled` opt-out continues to gate only the auto-open path.

### Manual send behavior

- Always uses the direct `UIApplication.open` path (no App Group keyboard handoff, since the screen is foregrounded in the main app).
- Resolves `{mode}` from the transcription's original `powerModeId`, falling back to the currently selected mode if that mode was deleted.
- Resolves `{preset}` from the active variation chip (nil for the original).
- **Uses `transcription.timestamp` for the note name**, not the tap time. With the default `VD {date} {HH}-{mm}-{ss}` template that means the note name is stable per transcription, so re-sending or sending a different variation of the same note appends into the same Obsidian file (the URL has `&append=true`) instead of producing a new file every tap.
- Bypasses the per-mode `obsidianEnabled` flag - explicit taps should override the auto-save opt-out.

## Test plan

- [ ] In Settings → Integrations, toggle "Show Send to Obsidian button" on; verify FAB appears on transcription detail screen.
- [ ] With both toggles off, no FAB shown and "Note name" field hidden.
- [ ] With only the new toggle on, "Note name" + footer still appear.
- [ ] Tap the Obsidian FAB on a multi-day-old recording; confirm Obsidian opens with the note name reflecting the recording date, not today.
- [ ] Generate a variation, switch to it via the chip bar, tap the FAB; verify the variation text lands in Obsidian and appends to the same file as the original (default template).
- [ ] Verify auto-open still works when "Auto-open after transcription" is on (regression check).
- [ ] Verify per-mode `obsidianEnabled` toggle in mode settings still suppresses auto-open but does NOT suppress the manual button.
- [ ] When both the Append-with-Voice FAB and the Obsidian FAB are enabled, both render side-by-side in the bottom-right safe area.


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #247, opened 2026-05-10
Add Export Settings screen and manual Export to Folder button

n0an:
## Summary

Mirrors the recent Obsidian split (PR #246) for the markdown folder-export integration.

### Settings reshape

Moved the inline `Section("Export")` out of the main Settings screen and into a new push screen `Settings -> Export Settings`. The new screen hosts:

- `Markdown Export` variation picker (top, applies to manual share + auto + button paths)
- `Auto-export after transcription` toggle - renamed from "Save markdown file", **storage key `isFolderExportGloballyEnabled` retained** so existing users keep their preference
- `Show Export to Folder button` toggle - new key `isFolderExportButtonEnabled`
- Folder picker + Clear folder (gated to appear when either toggle is on)

Removed the supporting state, `.fileImporter`, and folder-picker error alert from `SettingsView.swift`. Dropped `import UniformTypeIdentifiers` since it's no longer needed there. New `SettingsDestination.export` case wired to push `ExportSettingsView`.

### Manual export FAB

New `folderExportFloatingButton` on `TranscriptionDetailView`, gated by `isFolderExportButtonEnabled`. Renders as a third FAB in the bottom-right safe area, to the left of the Obsidian send button (when both are on).

Tap behavior:
- No folder picked → "Pick an export folder first" alert directing the user to Settings → Export Settings
- Folder picked → schedules a write via the new `FolderExportService.saveManually(transcription:)` entry point, gives a success haptic
- Disabled while recording, same as the other two FABs

### Service

- New `FolderExportService.saveManually(transcription:) -> Bool` - bypasses both the global auto-export toggle and the per-mode `folderExportEnabled` flag (explicit user action overrides the auto-save opt-out, same call we made for the manual Obsidian send). Reuses the existing bookmark resolution + write logic.

### Misc

- Small `HudViewDark` cleanup: uncomments a `.background(.black)` and changes it to `.opacity(0.1)`.

## Test plan

- [ ] Settings → Export Settings shows the new screen; row label and nav title both read \"Export Settings\".
- [ ] With both toggles off, only the Markdown Export picker shows; folder picker hidden.
- [ ] With either toggle on, folder picker appears; picking a folder + dismissing reflects the display name.
- [ ] Auto-export still works when \"Auto-export after transcription\" is on (regression check after rename).
- [ ] Per-mode `folderExportEnabled` toggle in mode settings still suppresses auto-export but does NOT suppress the manual button.
- [ ] Tap the Export FAB with no folder picked → alert points to Settings → Export Settings.
- [ ] Tap with a folder picked → check the picked folder for a new `VivaDicta-YYYY-MM-DD_HHmmss.md` file.
- [ ] When all three FABs (Export, Obsidian, Mic) are enabled, they render side by side without overflowing on iPhone SE width.

### DIFF
--- a/VivaDicta/Services/FolderExportService.swift
+++ b/VivaDicta/Services/FolderExportService.swift
@@ -70,28 +70,37 @@ enum FolderExportService {
         }
     }
 
+    /// Outcome of a manual export attempt. Distinguishes failures so the UI
+    /// can show specific guidance instead of a generic "something went wrong".
+    enum ManualSaveResult: Equatable {
+        case success(filename: String)
+        case noFolderPicked
+        case bookmarkUnusable
+        case writeFailed(String)
+    }
+
     /// Manually writes the transcription as a markdown file in the picked
     /// folder. Bypasses the global auto-export toggle and per-mode opt-out
     /// because the user explicitly tapped the Export to Folder button.
-    /// Returns whether a write was scheduled - callers can use this to
-    /// surface a "no folder picked" hint to the user.
+    ///
+    /// Awaits the actual write (unlike `saveIfEnabled`) so callers can
+    /// surface real success or failure to the user instead of giving a
+    /// false success signal when the bookmark is stale or access is revoked.
     @MainActor
-    @discardableResult
-    static func saveManually(transcription: Transcription) -> Bool {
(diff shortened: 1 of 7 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #249, opened 2026-05-13
Swift 6.2 concurrency cleanup: @concurrent adoption + redundant-await sweep

n0an:
## Summary

Modernization pass against Apple's April 2026 Swift Concurrency Q&A guidance and a code review against project skill `swift-concurrency-pro`. Eight concurrency-focused commits, no behavior changes intended, build clean throughout.

### What changed

- **Removed redundant `@MainActor in` from `Task` closures inside MainActor contexts** (16 sites). Under `SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor` + approachable concurrency, a `Task { ... }` created from a MainActor caller already inherits MainActor. Kept the annotation in 47 sites where the surrounding context is genuinely nonisolated (Darwin notification handlers, `WCSessionDelegate`, `Timer.scheduledTimer` callbacks, `BGTaskScheduler` `@Sendable` closures, `AVAudio` delegates).
- **Migrated all 11 `Task.detached` sites to `Task { @concurrent in }`**. Per Apple's guidance, `Task.detached` discards everything from the parent context (priority, locals, cancellation propagation). `@concurrent` only sheds isolation while keeping priority and task locals - the right idiom for "run off MainActor."
- **Set `SWIFT_STRICT_CONCURRENCY = complete` explicitly on main app, keyboard, and watch app**. With `SWIFT_VERSION = 6.2` this is already the implicit default - making it explicit alongside the other Swift concurrency flags.
- **Moved `@concurrent` from Spotlight call sites to the `AppState` methods themselves**. Six `Task { @concurrent in await appState.method(...) }` collapse to plain `Task { ... }` once the AppState methods are marked `@concurrent`. Intent now lives at the function definition instead of every call site. Also marked `removeTranscriptionFromSpotlight` `@concurrent` for symmetry.
- **Added debug-only main-thread assertions to `RAGIndexingService` `nonisolated` wrappers**. Apple's `nonisolated(nonsending)` default in Swift 6.2 means `nonisolated async` functions stay in the caller's isolation. The wrappers in this file pre-date that change and may now silently run embedding/vector work on MainActor when called from `@MainActor`. The assertions (`assert(!Thread.isMainThread)`) verify empirically and fire only in debug builds. If any assertion fires in normal use, the fix is to flip `nonisolated` to `@concurrent` on that wrapper.
- **Code-review follow-ups**: deleted unused `indexTranscriptionToSpotlight(_:Transcription)` and `updateTranscriptionInSpotlight(_:Transcription)` overloads in `AppState`; dropped redundant `@unchecked Sendable` from `ModelDownloadManager` (the class is implicitly `@MainActor` and therefore Sendable through actor protection); dropped redundant inner `[weak self]` in `LiveTranslationAudio.makeCaptureStream`'s nested `Task`.
- **Dropped 44 redundant `await` keywords** on synchronous `@MainActor` closure calls across `AppGroupCoordinator` (25 sites), `AIService`/`AIService+Chat` (10), and OAuth clients (9). These were the `'no async operations occur within await expression'` warnings that the explicit strict-concurrency setting surfaced.

Also includes a small UI tests target + launch-baseline added to the branch out-of-band; non-concurrency.

### Notes / Out of Scope

- **RAG `nonisolated` → `@concurrent` follow-up**: not done in this PR. The assertions are the verification step; if they fire in real use, the one-line fix lands in a follow-up.
- **`SonioxRealtimeSTTClient`/`TTSClient` reconnect race** (pre-existing): a stale `handleMessage` from an old `receiveTask` can yield into a new `AsyncStream` continuation if `connect()` is called before the previous `disconnect()`'s task has noticed cancellation. Not from this branch's work; flagged for a separate session-token fix.

## Test plan

- [x] `xcodebuild` clean build succeeds for VivaDicta scheme on iPhone simulator (iOS 26.4)
- [x] No new compiler warnings introduced
- [x] All 44 `'no async operations occur within await expression'` warnings eliminated (count went from 44 to 0)
- [x] Build clean across main app, VivaDictaKeyboard, VivaDictaWatch, VivaDictaWidget, ShareExtension, ActionExtension targets
- [ ] Manual smoke test: record a transcription, run AI enhancement (streaming), retranscribe, edit a variation, search Spotlight, open from Spotlight - all behave identically
- [ ] Manual smoke test: keyboard extension dictation flow + text processing flow unchanged
- [ ] Manual smoke test: Apple Watch recording transfer unchanged
- [ ] Manual smoke test: live translation start/stop cycle unchanged
- [ ] RAG assertion test: cold launch with Smart Search on + transcriptions in DB, observe whether any `RAG _X running on main thread` assertion fires


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Addressed the blocker both reviews flagged - commit 9d0c14cf9.

The RAG main-thread assertions were not passive checks: given `SWIFT_APPROACHABLE_CONCURRENCY = YES` (→ `nonisolated(nonsending)` default) plus `_chunkText` being synchronous, they were guaranteed to trap on the first Debug call from any `@MainActor` flow.

Fix: flipped all six async wrappers + `initializeLumoKit` from `nonisolated` to `@concurrent`. `_chunkText` was synchronous - `@concurrent` requires `async`, so it became `@concurrent async` and its two call sites (`RAGIndexingService.swift:324`, `:432`) now `await`. The assertions stay as genuine guard rails now, and the doc comment was rewritten to stop describing them as regression-detectors.

Build clean. Did not touch the two non-blocking cancellation-semantics notes (MainView zip / FolderExportService manual export) - those are behavior-preserving in practice; happy to add belt-and-suspenders cancellation checks if you'd prefer.

n0an:
@claude please re-review - blocker addressed in 9d0c14cf9

n0an:
@codex please re-review - blocker addressed in 9d0c14cf9

PR #254, opened 2026-05-15
Extract TranscriptionCore + CloudTranscription with OpenAI pilot

n0an:
## Summary

Splits the transcription layer into modules following Bev's `Repository -> Database + Networking + Domain` pattern. This is **PR B1** from the modularization plan (`llmtemp/modularization-plan-2026-05-15.md`).

### New modules

**`Modules/TranscriptionCore`** - pure types, no app-target deps
- `TranscriptionServiceResult`
- `TranscriptionError`
- `TranscriptionProgressInfo` + `TranscriptionProgressStage` + `TranscriptionProgressHandler`

Justified as a standalone module because two real modules will depend on it (`CloudTranscription` now, `LocalTranscription` in PR B3).

**`Modules/CloudTranscription`** - first real cloud impl in a module
- `CloudTranscriptionError` (moved from the in-app router file)
- `NetworkRetry` (cross-module isolation via `isolation: isolated (any Actor)? = #isolation` so existing callers keep their non-Sendable captures)
- `URL.audioMIMEType` extension
- `OpenAITranscriptionService` refactored to take `Config(apiKey, modelName, language)` via init. Stateless apart from its config; callers build one per request.
- `CloudTranscriptionMocks` library exports `MockTranscriptionService` (Bev-style stub / call count / didCall mock).

**`Modules/LocalTranscription`** - empty placeholder for PR B3 (WhisperKit). Not yet linked to any target.

### App-target changes

- `CloudTranscriptionService` router still owns the 11 unmigrated services; the `.openAI` case now builds the new module's service inline with a `Config`, pulling language from `UserDefaultsStorage` at call time.
- 18 app files gained `import TranscriptionCore`; 11 cloud service files also gained `import CloudTranscription` for the shared error/retry helpers.
- Net **-326 LOC** in the app target (types and one impl moved out, no logic deleted).

### Why not move the protocol or `TranscriptionModel` in this PR

`TranscriptionService` protocol still has a `model: any TranscriptionModel` parameter, and `TranscriptionModelProvider` (801 LOC of model catalogs) is deeply coupled to app-target types like `AIProvider`, `CloudModel.apiKey`, `CustomTranscriptionModel`. Moving those is a separate refactor; doing it here would have doubled the PR size for no immediate test or build-time win. Future PRs (B2-B5) can promote the protocol once more services share the init-based config pattern.

## Test plan

- [x] `swift test --package-path Modules/TranscriptionCore` - 7 tests pass
[...]

PR #255, opened 2026-05-15
B2: Migrate all 11 cloud transcription services into CloudTranscription module

n0an:
## Summary

**PR B2** from the modularization plan (`llmtemp/modularization-plan-2026-05-15.md`). Completes the cloud transcription extraction started in PR #254 (OpenAI only). All 11 cloud providers + the Custom OpenAI-compatible endpoint client now live in `Modules/CloudTranscription`, each with a per-service `Config` struct.

### Services moved this PR

**First commit (`ebed002ae`)** - 4 simple services that only need `(apiKey, modelName, language)`:
- Gemini, ElevenLabs, Cohere, Cartesia

**Second commit (`8f72f4e26`)** - 6 harder services + Custom + a shared utility:
- **Groq** - adds `vocabulary: [String]` (folded into the OpenAI-compatible `prompt` field)
- **Deepgram** - adds `vocabulary` + `isSpeakerDiarizationEnabled`. Router still maps the `nova-3-multilingual` alias to `(modelName: "nova-3", language: "multi")` before constructing the config
- **Mistral** - adds `isSpeakerDiarizationEnabled`. Public static `requestLanguage(for:diarizationEnabled:)` preserves the quirk that diarization disables explicit language hints
- **Soniox** - upload + create + poll flow with `vocabulary`, `isSpeakerDiarizationEnabled`, `translationTargetLanguage`
- **Gladia** - same shape. Config omits `modelName` because the v2 pre-recorded endpoint ignores it
- **Speechmatics** - same shape. Service internally maps `zh` -> `cmn`
- **CustomTranscriptionService** - decoupled from app-target `CustomTranscriptionModel`; router builds the Config from its stored properties
- **SpeakerDiarizationFormatter** + `SpeakerTurn` promoted to `TranscriptionCore` so both cloud and future local services share it. WhisperKit (still in app target) already imports `TranscriptionCore`, so it picks up the new home automatically.

### App-target changes

- `CloudTranscriptionService` router is now ~150 LOC: pulls language/diarization/translation/vocabulary from `UserDefaultsStorage` + `AppGroupCoordinator` + `CustomVocabulary`, builds the right `Config` per case, calls the service. No more `lazy var` instances - each call constructs a fresh stateless service.
- 11 service files + `SpeakerDiarizationFormatter.swift` deleted from `VivaDicta/Services/Transcription/`.
- Net **-2318 LOC** in the app target across both commits.

### Pattern (per-service Config) vs alternative (shared Config)

I went with per-service `Config` structs for two reasons:
1. Consistent with the OpenAI pattern established in PR #254.
2. Each service declares exactly what it consumes - no optional fields nobody uses, no "unused field" smell. Cohere/Cartesia get `(apiKey, modelName, language)`; Soniox/Gladia/Speechmatics get the full diarization + translation set.

The downside is the router has to build the right Config per case. That's fine because the router already had a per-provider switch.

### Language normalization moved to the router

For Cohere and Cartesia, "auto" or unsupported codes fall back to "en". That validation used to live inside the service (which reached for `TranscriptionModelProvider.cohereLanguages.keys`). Now the router does the normalization with the same supported-code set, and the service treats `language` as a ready-to-send opaque value. Clean separation.

## Test plan

- [x] `swift test --package-path Modules/CloudTranscription` - 9 tests pass
- [x] `swift test --package-path Modules/TranscriptionCore` - 7 tests pass
- [x] iOS app build green
- [ ] CI verifies the full app test suite
- [ ] Smoke: actually transcribe with each provider that has an API key configured (Codex review will likely flag this is the place real bugs would hide)

## Risk

The diarization/translation paths are not unit-tested. Each service was carefully refactored 1:1 against the original. Any regression would show up as runtime behavior change (wrong/missing transcription, wrong language sent to API, etc.) - the kind of thing that needs live API testing to catch.


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #262, opened 2026-05-16
Relocate module-tests from app target into their owning modules

n0an:
## Summary

Six test files sat in `VivaDictaTests/` but exercised types living entirely inside SPM modules. Moved them to where their SUTs live so the test pyramid matches the production layering.

### Moves

| File | From | To | Tests |
|------|------|-----|-------|
| `AppGroupCoordinatorSessionTests.swift` | `VivaDictaTests/` | `Modules/AppGroup/Tests/AppGroupTests/` | 9 |
| `AppGroupCoordinatorStateTests.swift` | `VivaDictaTests/` | `Modules/AppGroup/Tests/AppGroupTests/` | 13 |
| `SpeakerDiarizationFormatterTests.swift` | `VivaDictaTests/` | `Modules/TranscriptionCore/Tests/TranscriptionCoreTests/` | 3 |
| `PresetCatalogTests.swift` | `VivaDictaTests/` | `Modules/Presets/Tests/PresetsTests/` | ~10 |
| `PresetManagerTests.swift` | `VivaDictaTests/` | `Modules/Presets/Tests/PresetsTests/` | ~25 |
| `PresetTests.swift` | `VivaDictaTests/` | `Modules/Presets/Tests/PresetsTests/` | ~12 |

### What changed in each file

- `git mv` to preserve history
- Dropped the dead `@testable import VivaDicta` line (no test referenced any app-target type - it was leftover from when these files predated the module extractions)
- Changed `import <Module>` to `@testable import <Module>` to keep access to internals consistent with the existing module-test pattern

### Results

- App test bundle: 21 → 15 files
- Each module's package tests now own their SUT's tests
- All 87 relocated tests pass in their new homes
- No production code touched

## Test plan
- [x] `swift test --package-path Modules/AppGroup` → 22 tests pass
- [x] `swift test --package-path Modules/TranscriptionCore` → 10 tests pass (3 newly added + 7 existing)
- [x] `swift test --package-path Modules/Presets` → 55 tests pass (47 newly added + 8 existing)
- [x] `xcodebuild build` for the main app passes

PR #266, opened 2026-05-16
Wire all 10 module test targets into the Xcode test plan

n0an:
## Summary

Closes a long-standing CI gap: 380+ tests across our 10 SPM module test bundles were silently skipped on `unit-tests.yml` (Codex flagged this on PR #262).

### The problem

CI runs `xcodebuild test -scheme VivaDicta`, which executes whatever the scheme's test plan (`VivaDictaTestPlan.xctestplan`) lists. The plan only included:

- `VivaDictaTests` (242 tests)
- `VivaDictaUITests` (disabled)

Every module test bundle - 11 of them, accumulating 160+ tests we've added across this and prior sessions - was unreachable from the main scheme.

### The fix

Two changes:

**1. Add all 10 module test targets to the test plan**

Each module's `*Tests` target is now a top-level entry in `VivaDictaTestPlan.xctestplan`:

- `AppGroupTests` (22 tests)
- `CloudTranscriptionTests` (33 tests)
- `KeychainTests` (6 tests)
- `LocalTranscriptionTests` (10 tests)
- `NetworkingTests` (5 tests)
- `OAuthTests` (11 tests)
- `PresetsTests` (55 tests)
- `TranscriptionCoreTests` (10 tests)
- `TranscriptionKitTests` (5 tests)
- `TestUtilitiesTests` (3 tests)

Added through Xcode UI; reflected in the .xctestplan JSON.

**2. Add `Networking` as a direct dep of `TranscriptionKit`**

Running the full test plan exposed a hidden link issue: `TranscriptionKit.o` references the `URLSession: URLSessionProtocol` witness table from Networking, because `TranscriptionEngine` calls `OpenAITranscriptionService(config:)` (and other cloud services) using their default `urlSession: any URLSessionProtocol = URLSession.shared`. That default arg inlines a witness-table reference at the call site.

SPM doesn't auto-propagate this transitive conformance dep through the umbrella in every build context. The test-plan context specifically fails to link without an explicit dep on Networking from TranscriptionKit. Adding it explicitly resolves the link.

See the inline comment added in `TranscriptionKit/Package.swift`.

### After this PR

`xcodebuild test -scheme VivaDicta` now executes **all 402 tests across 11 test bundles**. Same for ⌘U in Xcode locally. CI's `unit-tests.yml` doesn't need any changes - it just calls the same command.

## Test plan
- [x] `xcodebuild test -scheme VivaDicta -workspace ./VivaDicta.xcodeproj/project.xcworkspace -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max,OS=26.4'` passes: 402 tests across 11 bundles, 0 failures

### DIFF
--- a/Modules/TranscriptionKit/Package.swift
+++ b/Modules/TranscriptionKit/Package.swift
@@ -15,6 +15,7 @@ let package = Package(
         .package(path: "../TranscriptionCore"),
         .package(path: "../CloudTranscription"),
         .package(path: "../LocalTranscription"),
+        .package(path: "../Networking"),
     ],
     targets: [
         .target(
@@ -23,6 +24,13 @@ let package = Package(
                 .product(name: "TranscriptionCore", package: "TranscriptionCore"),
                 .product(name: "CloudTranscription", package: "CloudTranscription"),
                 .product(name: "LocalTranscription", package: "LocalTranscription"),
+                // Direct dependency on Networking required so the linker can
+                // resolve the `URLSession: URLSessionProtocol` witness table
+                // referenced by CloudTranscription service initializers with
+                // their default `urlSession: URLSessionProtocol = URLSession.shared`.
+                // SPM doesn't propagate this transitive conformance link
+                // through the umbrella in every build context.
+                .product(name: "Networking", package: "Networking"),
             ]
         ),
         .testTarget(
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #265, opened 2026-05-16
Move Color+Ext and View+Ext into DesignSystem; link to Keyboard target

n0an:
## Summary

Second `DesignSystem` PR. Moves the two remaining `Utils/Extensions/*.swift` files into the module.

### What moves

- **`Color+Ext.swift`** - small. `Color.random()`, `var sui: Color`.
- **`View+Ext.swift`** - the bigger surface:
  - `BlurTransition`, `Transition.blur(radius:)` static
  - `BadgeStyle` protocol + `FancyBadgeStyle` impl + `OverlayBadge` modifier + `.badge(...)` + `.badgeStyle(...)` + `.fancy` shorthand
  - Glass modifiers: `glassEffectOrMaterial`, `glassEffectClear`, `glassEffectColor`, `glassDismissCircle`, `glassFABCircle`, `glassCapsule`, `modelCardBackground`
  - Button styling: `prominentButton`, `animatedCopyButtonStyle`, `RecordButtonButtonStyle`
  - Misc: `iflet`, `onFirstAppear`, `minimizedSearch`, `debugBorder`

All public-API extension methods marked `public`. Implementation-detail modifier structs (e.g. `GlassEffectOrMaterialModifier`, `ModelCardBackgroundModifier`) remain internal - the public surface is the `View` extension methods, not the modifiers themselves.

### Wiring

24 consumer files get `import DesignSystem`:

- **18 in main app target** - cards, views, settings screens (e.g. `MainView`, `TranscriptionDetailView`, `LocalModelCard`, `ModeEditView`, `WhatsNewContent`, etc.)
- **6 in `VivaDictaKeyboard` extension** - `FullAccessPromptView`, `RecordingStateView`, `ProcessingStateView`, `TextProcessingStateView`, `RewriteModesView`, `KeyboardViewController`

`VivaDictaKeyboard` target gains `DesignSystem` as a package product dependency (added via Xcode UI; reflected in `project.pbxproj`).

### pbxproj cleanup

Removes two stale references from `membershipExceptions`:
- `VivaDictaWidgetExtension`'s exception set drops `Utils/Extensions/Color+Ext.swift` and `Utils/Extensions/View+Ext.swift` (these were verbatim file inclusions via folder-sync; Widget didn't actually use any of the APIs, so no follow-up wiring needed for that target)
- `VivaDictaKeyboard`'s exception set drops the same two entries (Keyboard does use the APIs - now consumed via `import DesignSystem`)

### Platform tweak

[...]

PR #269, opened 2026-05-16
TagFilterBar: use DesignSystem.glassCapsule instead of private modifier

n0an:
## Summary

`TagFilterBar.swift` had a private `ChipBackgroundModifier` that was a byte-for-byte duplicate of the `glassCapsule(tint:fallback:)` helper in DesignSystem - same iOS 26 glass effect with optional tint, same capsule-shaped fallback. Removed the local modifier in favor of the DS helper.

Net: -25 LOC (private modifier + view extension deleted), file gains `import DesignSystem`.

Visual fidelity preserved:
- Dark-mode `color.opacity(0.6)` adjustment factored into a local `chipColor` constant inside `chipButton(...)`.
- Compound `.opacity(0.2)` fallback factored into `chipFallback` constant.
- Both feed `glassCapsule`'s `tint` and `fallback` parameters identically to the previous implementation.

## Why this matters

This is the first real-app consumer of `glassCapsule` since it landed in DesignSystem. Claude flagged the duplication in PR #265's review. Now DS earns its keep in TagFilterBar instead of just being declared next to it.

## Test plan
- [x] `xcodebuild build` for the main app passes
- [ ] Open Tags filter bar in light + dark mode, verify chip backgrounds render identically to before
- [ ] Tap a chip to select; confirm tint applies correctly on iOS 26

### DIFF
--- a/VivaDicta/Views/SettingsScreen/Tags/TagFilterBar.swift
+++ b/VivaDicta/Views/SettingsScreen/Tags/TagFilterBar.swift
@@ -6,6 +6,7 @@
 //
 
 import SwiftUI
+import DesignSystem
 
 /// Horizontal scrollable filter bar with source tags and user tags.
 struct TagFilterBar: View {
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #273, opened 2026-05-18
Remove 35 redundant Task.checkCancellation() calls

n0an:
## Summary

Sweep of `Task.checkCancellation()` calls that sit immediately next to another cancellation point and therefore add no real cancellation responsiveness - just line noise.

53 -> 17. Build verified with `xcodebuild`.

### Removed

- **Before `URLSession.shared.data/bytes(for:)`** (17 calls) - URLSession's async methods are cancellation-aware and throw on cancel.
- **After `URLSession.shared.data(...)` before sync JSON parse** (4 calls) - the only intervening work is microseconds of parsing.
- **Inside `for try await line in bytes.lines`** (8 calls) - URLSession byte stream is cancellation-aware via its async iterator.
- **Before `try await Task.sleep(...)`** in cloud transcription poll loops (3 calls) - `Task.sleep` itself throws `CancellationError` on the first suspend.
- **Top of poll loop** in `RecordAndReturnTranscriptionIntent` (1 call) - the same loop sleeps at the bottom, which already catches cancel.
- **Between two cancellation-aware async calls** in `LocalModelDownloader.downloadParakeet` (1 call).
- **Stale doc comment** referencing the removed check.

### Kept (load-bearing or defensible)

- CPU-bound chunked loops in `ParakeetTranscriptionService` (canonical use of explicit checks)
- Continuation boundaries in `KeyboardTextProcessor` - `withCheckedThrowingContinuation` does not auto-propagate cancellation
- Phase boundaries in `RecordViewModel` (downsample / transcribe / enhance)
- State-machine guards in `ModelDownloadManager` (prevents "downloaded" state on cancel race)
- First-iteration check in `NetworkRetry.withRetry` (only cancellation point before the first operation runs)
- WhisperKit phase boundaries in `LocalModelDownloader` - notably the post-loadModels check exists so unloadModels can still run for memory cleanup
- UI state safety in `TranscriptionDetailView.extractReminderSuggestions`

## Test plan
- [x] `xcodebuild -scheme VivaDicta build` succeeds
- [ ] Spot-check cancel behavior: cancel an in-flight AI enhancement and verify the task stops promptly (URLSession-driven cancel)
- [ ] Spot-check cancel during cloud transcription poll loop (Speechmatics/Soniox/Gladia)
- [ ] Spot-check cancel during model download (Parakeet/WhisperKit)

### DIFF
--- a/Modules/CloudTranscription/Sources/CloudTranscription/GladiaTranscriptionService.swift
+++ b/Modules/CloudTranscription/Sources/CloudTranscription/GladiaTranscriptionService.swift
@@ -207,7 +207,6 @@ public struct GladiaTranscriptionService: TranscriptionService, Sendable {
                 throw CloudTranscriptionError.apiRequestFailed(statusCode: 504, message: "Transcription timed out")
             }
 
-            try Task.checkCancellation()
             try await Task.sleep(for: .nanoseconds(pollIntervalNanoseconds))
         }
     }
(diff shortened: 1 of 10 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #285, opened 2026-05-24
Extract CustomOpenAIService from AIService (Phase 2a, provider #2)

n0an:
Second incremental step of the AIService decomposition (Phase 2 of the architecture migration plan). Pulls the Custom OpenAI endpoint verification logic - HTTP probe + status-code-to-message mapping - out of AIService into a self-contained struct.

## 🟢 Stealth bug fix (call-out from Claude's review)

The original \`testCustomOpenAIEndpoint\` had \`catch let error as URLError\` before the generic catch, but \`DefaultNetworkService.callData\` wraps every URLError as \`NetworkError.transport(error)\`. The typed catch arm therefore **never matched in production** - every transport error fell through to the generic \`"Error: ..."\` NSError message. Users never saw the friendly \"Cannot connect to server\", \"Connection timed out\", \"No internet connection\" strings.

The new \`CustomOpenAIService.unwrapURLError\` unwraps \`NetworkError.transport(URLError)\` properly, so these friendly messages will actually surface for the first time. This is a behavior improvement on top of the refactor.

## What moves

- **\`CustomOpenAIService.swift\`** (new) in \`VivaDicta/Services/AIEnhance/Providers/\`:
  - \`testEndpoint(endpointURL:modelName:apiKey:)\` - sends a minimal chat-completions POST and maps HTTP status / URLError codes into a structured \`TestResult\` (\`.success(message:)\` / \`.failure(message:)\`). Never throws.
  - \`TestResult\` enum carries a user-readable message for both outcomes.
  - \`unwrapURLError(_:)\` - extracts the underlying URLError from a raw URLError throw, a \`NetworkError.transport(URLError)\` wrap, or a \`String(describing:)\` fallback (see "Two paths" below).

- **\`AIService.verifyCustomOpenAISetup()\`** shrinks from ~100 lines to ~20: validation (empty URL / malformed URL / empty model name) stays here because it reads \`@Observable\` config state; HTTP probe delegates to \`CustomOpenAIService.testEndpoint\`. Translates the \`TestResult\` to the existing \`(Bool, String)\` tuple shape the UI consumes.
- **\`AIService.testCustomOpenAIEndpoint()\`** deleted - now lives in the new service.

AIService loses ~95 lines net (2,960 → ~2,865). State ownership stays with AIService.

## Two paths, two contexts

\`unwrapURLError\` has a typed-cast clause for production and a \`String(describing:)\` regex fallback for the test bundle:

- **In production**, \`error as? NetworkError\` succeeds and the helper returns via the typed-NetworkError-unwrap clause.
- **In the test bundle**, the typed cast returns \`nil\` even when the dynamic type IS \`NetworkError\` - apparently because \`Networking\` is loaded twice (once into the app, once via the test target's explicit \`import Networking\` alongside \`@testable import VivaDicta\`), producing two distinct \`NetworkError\` types with the same nominal name. The regex fallback parses the bridged NSError description (\`transport(Error Domain=NSURLErrorDomain Code=-1004 ...)\`) to recover the URLError code.

If the dual-link is ever fixed (e.g., by making \`Networking\` a dynamic library), the fallback becomes unreachable - but it's cheap insurance.

## Tests

\`CustomOpenAIServiceTests\` (16 tests) in \`VivaDictaTests/\`:
- Happy path: 200 → success message
- Request shape: Bearer auth, minimal body with model name
- API key handling: nil / empty / non-empty
- Status mapping: 401, 404, 400-with-JSON-error, 400-without, 5xx
- Transport errors via three stub forms:
  - \`NetworkError.transport(URLError)\` (exercises the regex fallback in the test bundle)
  - \`URLError\` thrown directly (production-reachable path)
  - \`NSError(domain: NSURLErrorDomain, ...)\` (bridges to URLError, matches the URLSession-layer reality)

## What's NOT in this PR

- The Custom OpenAI chat / enhance branches in AIService's big switch statements still live in AIService. Sub-phase 2b will swallow those.
- No \`AIProvider\` protocol yet. Two extracted (\`OllamaService\`, \`CustomOpenAIService\`); one more (\`Apple FM\`) completes sub-phase 2a, then the protocol shape becomes clear.

## Test plan

- [x] \`swift test\` in Modules/Networking: 23 tests pass
- [x] \`swift test\` in Modules/OAuth: 15 tests pass
- [x] \`swift test\` in Modules/CloudTranscription: 50 tests pass
- [x] \`xcodebuild test\` for VivaDicta scheme: 45 tests pass (was 29; +16 \`CustomOpenAIServiceTests\`) on iPhone 17 Pro Max / iOS 26.4

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -1992,7 +1992,6 @@ class AIService {
         }
     }
 
-    /// Verifies Custom OpenAI setup and returns status message
     /// Validates the Custom OpenAI configuration and probes the endpoint.
     /// Validation stays here (we read the `@Observable` config state);
     /// the HTTP probe delegates to `CustomOpenAIService`.
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -1993,124 +1993,30 @@ class AIService {
     }
 
     /// Verifies Custom OpenAI setup and returns status message
+    /// Validates the Custom OpenAI configuration and probes the endpoint.
+    /// Validation stays here (we read the `@Observable` config state);
+    /// the HTTP probe delegates to `CustomOpenAIService`.
     public func verifyCustomOpenAISetup() async -> (success: Bool, message: String) {
         let endpointURL = customOpenAIRequestURL
         let modelName = customOpenAIModelName
 
-        // Check URL is configured
         guard !endpointURL.isEmpty else {
             return (false, "Endpoint URL is not configured")
         }
-
-        // Validate URL format
         guard URL(string: endpointURL) != nil else {
             return (false, "Invalid endpoint URL format")
         }
-
-        // Check model name is configured
         guard !modelName.isEmpty else {
             return (false, "Model name is not configured")
         }
 
-        // Test connection with a minimal chat completions request
-        return await testCustomOpenAIEndpoint()
-    }
-
-    /// Tests the Custom OpenAI endpoint with a minimal request and returns detailed status
-    private func testCustomOpenAIEndpoint() async -> (success: Bool, message: String) {
-        let endpointURL = customOpenAIRequestURL
-        let modelName = customOpenAIModelName
-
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #286, opened 2026-05-24
Extract AnthropicService from AIService (Phase 2a, provider #3)

n0an:
## Summary

- Third provider extraction in Phase 2a of the AIService decomposition. Pulls Anthropic's three pure-HTTP call sites (non-streaming enhance, streaming enhance, API-key verification) into a new \`AnthropicService\` struct in \`VivaDicta/Services/AIEnhance/Providers/\`.
- Anthropic is NOT OpenAI-compatible: it uses \`/v1/messages\` with a different body shape (\`system\` is top-level, \`max_tokens\` is required, headers use \`x-api-key\` + \`anthropic-version\` instead of Bearer auth) and different SSE event format (\`content_block_delta\` events with \`delta.text\`, not OpenAI's \`choices[].delta.content\`). The new struct makes that divergence visible at the file boundary.
- AIService loses 134 lines net (2896 → 2762). State (connection flags, mode/model selection, CLI Server routing via VivAgentsClient) stays in AIService; the new struct is pure HTTP and stateless apart from injected dependencies.

## What's new

- \`Providers/AnthropicService.swift\` (~220 lines): \`enhance\`, \`enhanceStreaming\`, \`verifyAPIKey\`. Inline trim + \`AIEnhancementOutputFilter\` so the struct stays self-contained (no callback into AIService).
- \`AnthropicServiceTests.swift\` (14 tests): request-shape verification (x-api-key vs Bearer, anthropic-version header, system top-level, max_tokens required), status-code mapping (200/429/5xx/other), malformed body, verifyAPIKey behavior, streaming request shape.

## Streaming testing limitation

Same caveat as \`OllamaService\` / \`CustomOpenAIService\`: \`MockNetworkService.bytes(for:)\` can't return a real \`URLSession.AsyncBytes\` because the type has no public initializer. Streaming tests verify the outgoing request shape (headers, body, stream flag) but the SSE parser itself relies on integration / manual testing.

## Test plan

- [x] \`xcodebuild test -only-testing:VivaDictaTests/AnthropicServiceTests\` - all 14 pass
- [x] Full app build succeeds
- [x] Full test suite: \`AnthropicServiceTests\` green; the pre-existing flaky \`AIServiceNetworkingTests/defaultNetworkServiceIsDefaultNetworkServiceWhenNotInjected\` (dual-module-link issue documented in #285's \`CustomOpenAIService\` unwrapURLError) reproduces on \`main\` too and is unrelated to this change

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -1132,82 +1132,14 @@ class AIService {
         model: String,
         onPartialResponse: @escaping @MainActor (String) -> Void
     ) async throws -> String {
-        var request = URLRequest(url: URL(string: AIProvider.anthropic.baseURL)!)
-        request.httpMethod = "POST"
-        request.addValue("application/json", forHTTPHeaderField: "Content-Type")
-        request.addValue("text/event-stream", forHTTPHeaderField: "Accept")
-        request.addValue(apiKey, forHTTPHeaderField: "x-api-key")
-        request.addValue("2023-06-01", forHTTPHeaderField: "anthropic-version")
-        request.timeoutInterval = baseTimeout
-
-        let requestBody: [String: Any] = [
-            "model": model,
-            "max_tokens": 8192,
-            "system": systemMessage,
-            "messages": [
-                ["role": "user", "content": userMessage]
-            ],
-            "stream": true
-        ]
-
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #290, opened 2026-05-24
Add Save options bottom sheet to Live Translation

n0an:
## Summary
- Tapping **Save as note** in Live Translation now opens a bottom sheet (`.medium` detent) instead of always saving the combined source + translation
- User picks one of three options: **Both languages**, **{Source} only**, or **{Target} only**
- Options with no content are hidden, so the user can't save an empty section
- "Original only" / "Translated only" save the raw text without language headings; "Both" keeps the previous heading layout

## Test plan
- [ ] Open Live Translation, run a session in two languages, tap Stop
- [ ] Tap **Save as note** - verify a bottom sheet appears with all three options
- [ ] Pick **Both languages** - verify the note contains both sections with language headings
- [ ] Repeat with **{Source} only** - verify the note contains only the source text, no headings
- [ ] Repeat with **{Target} only** - verify the note contains only the translation, no headings
- [ ] Run a session where translation is empty (e.g. stop immediately) - verify the translated-only and "Both" options are hidden
- [ ] Tap Cancel - verify nothing is saved

### DIFF
--- a/VivaDicta/Views/LiveTranslation/LiveTranslationView.swift
+++ b/VivaDicta/Views/LiveTranslation/LiveTranslationView.swift
@@ -75,11 +75,14 @@ struct LiveTranslationView: View {
                 )
             }
             .sheet(isPresented: $showingSaveOptions) {
-                SaveOptionsSheet(snapshot: service.transcriptSnapshot()) { content in
+                // Capture the snapshot once here so the option-visibility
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #292, opened 2026-05-25
Align TestUtilities Evaluate + port SwiftExpectation / ObservationTestUtils

n0an:
## Summary

Trims `Optional<Result<T, Error>>.evaluate()` down to the canonical Swift Testing pattern (`Issue.record` + `throw`) and drops the custom autoclosure message and source-location plumbing that was pure ceremony. `StubNotSetError` stays public because `MockNetworkService` / `MockURLSession` throw it directly for non-`evaluate()` test-infra failures (mismatched stub type, fundamentally unstubbable methods).

Ports two more helpers from the same upstream:
- **`SwiftExpectation`** - Swift Testing replacement for `XCTestExpectation`. Manual `fulfill()` + `wait(timeout:)`.
- **`ObservationTestUtils.changes(to:on:trigger:)`** - waits for an `@Observable` property to change before asserting. Uses a **trigger closure** form (instead of the upstream `async let` pattern) because non-`Sendable` `@Observable` types trip Swift 6 strict concurrency when sent across actor boundaries. Trigger form stays on one actor end-to-end.

Real usage proves both:
- `PresetManagerTests.addPreset_firesObservationOnPresets` and `deletePreset_firesObservationOnPresets` use `changes(to: \.presets, on: sut, ...)` to verify observation fires, followed by explicit `#expect` assertions for the data effect.
- These tests transitively exercise `SwiftExpectation` (used inside `changes()`).

Also hoists `sut` to a property + `init()` in `PresetManagerTests` since the SUT config is constant across tests. Two persistence tests stay inline because they intentionally build manager pairs against the same defaults to verify cross-instance reads.

## Notes for reviewers

- **Deviation from upstream pattern**: upstream `changes(to:on:)` returns immediately after observation setup and pairs with `async let` in the caller. That doesn't compile under Swift 6 strict concurrency for non-`Sendable` `@Observable` types (which `PresetManager` is). Switched to a trailing `trigger:` closure - same shape (setup → mutate → wait), single actor.
- **`@MainActor` on the two observation tests**: the helper is `@MainActor` because the trigger closure captures a non-`Sendable` `@Observable`. Tests need to be `@MainActor` to call it without sending the observable across actors. SPM package targets don't have `SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor` like the Xcode project does, so the annotation is explicit. Setting `defaultIsolation` on the Presets target could remove these annotations later - deferred.
- **Bev-pattern reference**: this aligns several test-utility patterns with the structure used in https://github.com/jacobbartlett/Bev. The trigger-closure form is our adaptation; otherwise the helpers match.

## Test plan

- [x] `swift test` in `Modules/Presets/` - 57/57 pass including the two new observation tests
- [x] `swift test` in `Modules/TestUtilities/` - 3/3 pass
- [x] `xcodebuild build` for VivaDicta - SUCCEEDED
- [ ] Confirm full xcodebuild test suite still passes (pre-existing `AIServiceNetworkingTests` parallel flake unrelated to this PR)

### DIFF
--- a/Modules/Presets/Package.swift
+++ b/Modules/Presets/Package.swift
@@ -12,6 +12,9 @@ let package = Package(
         .library(name: "Presets", targets: ["Presets"]),
         .library(name: "PresetsMocks", targets: ["PresetsMocks"]),
     ],
+    dependencies: [
+        .package(path: "../TestUtilities"),
+    ],
     targets: [
         .target(
             name: "Presets"
@@ -22,7 +25,11 @@ let package = Package(
         ),
         .testTarget(
             name: "PresetsTests",
-            dependencies: ["Presets", "PresetsMocks"]
+            dependencies: [
+                "Presets",
+                "PresetsMocks",
+                .product(name: "TestUtilities", package: "TestUtilities"),
+            ]
         ),
     ],
     swiftLanguageModes: [.v6]
--- a/Modules/Presets/Tests/PresetsTests/MockPresetSyncTests.swift
+++ b/Modules/Presets/Tests/PresetsTests/MockPresetSyncTests.swift
@@ -12,9 +12,14 @@ import PresetsMocks
 
 @Suite("MockPresetSync contract")
 struct MockPresetSyncingTests {
+    
+    var sut: MockPresetSync
+    
+    init() {
+        self.sut = MockPresetSync()
+    }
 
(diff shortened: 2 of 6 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #294, opened 2026-05-25
Clean up Swift Testing macro usage and fix UserDefaults suite race

n0an:
## Summary
- Replaced 80 `#expect(!X)` calls with `#expect(X == false)` across 20 test files. The leading `!` defeats Swift Testing's macro expansion - on failure the macro only sees the final negated boolean and reports an opaque \"expression was false\" instead of surfacing the actual operand. Rewriting as `X == false` lets the macro introspect both sides and print the real value when a test fails.
- Fixed a parallel-execution race in `AudioCleanupServiceTests` and `NoteCleanupServiceTests`: both used a fixed UserDefaults suite name (`\"AudioCleanupServiceTests\"`, `\"NoteCleanupServiceTests\"`) and called `removePersistentDomain` at the top of each test. Under Swift Testing's default parallel execution, one test could wipe another test's defaults mid-run. Suffixed the suite name with `UUID().uuidString` so each test gets its own private suite.

No behavior changes - the assertions still pass/fail on the same conditions. This is purely a diagnostic-quality and test-isolation improvement.

## Test plan
- [x] `xcodebuild ... build` - 0 errors
- [x] `xcodebuild ... build-for-testing` - 0 errors
- [ ] Full test run on CI

### DIFF
--- a/Modules/AudioRecording/Tests/AudioRecordingTests/AudioRecordingTests.swift
+++ b/Modules/AudioRecording/Tests/AudioRecordingTests/AudioRecordingTests.swift
@@ -16,7 +16,7 @@ struct DefaultAudioFileServiceTests {
 
         try sut.move(from: source, to: destination)
 
-        #expect(!FileManager.default.fileExists(atPath: source.path))
+        #expect(FileManager.default.fileExists(atPath: source.path) == false)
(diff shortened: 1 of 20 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #295, opened 2026-05-25
Hoist SUT, parameterize repetitive tests, modernize throws assertions

n0an:
Three independent test-only refactors bundled together since they all came out of the same Swift Testing audit.

### 1. Hoist SUT to property + init()

Where every test in the file built the same SUT, moved it to a struct property initialized in `init()`. Renamed the local to `sut` per the project convention. With Swift Testing's per-test struct instantiation each test still gets a fresh SUT — this just removes the duplication.

- `TranscriptionManagerTests` — `mockEngine` + `sut` hoisted (4 tests)
- `TranscriptionManagerConfigTests` — `sut: TranscriptionManager` hoisted, `manager` → `sut` (7 tests)
- `DataControllerTests` — `sut: DataController` hoisted with throwing `init()` (drops a `try!`), `controller` → `sut`, `makeDataController()` helper removed (16 tests)

### 2. Parameterize language-filler tests

`TextProcessingFilterTests` had 5 near-identical `filter_<lang>Fillers_removedWhenLanguageIs<Lang>` tests differing only by `(language, input, removedTokens, keptTokens)`. Collapsed into a single `@Test(arguments: [...])` over a tuple. Test count is preserved (each tuple is a separate case in test reports), and a failure now gets a useful `[lang] expected \"token\" to be removed; got: ...` message that names which row failed.

### 3. Modernize throws assertions

Replaced `do/catch + Issue.record` with `#require(throws:) + guard case` in the AI-provider service tests. The Swift 6.1+ `#require(throws:)` returns the captured error, so we can use a single `guard case` for the case match instead of a nested `if case` inside the `catch`. Also removes the bottom-of-block `catch { Issue.record(...) }` since `#require(throws:)` already fails the test when nothing or the wrong type is thrown.

- `AnthropicServiceTests` — 3 blocks
- `OpenAICompatibleServiceTests` — 4 blocks (the `enhanceRethrowsURLErrorTransport` test got the biggest cleanup)
- `XaiTranscriptionServiceTests` — 2 blocks
- `GroqTranscriptionServiceTests` — 2 blocks
- `OpenAITranscriptionServiceTests` — 2 blocks

`NetworkErrorBridgeTests` was checked but already uses the clean `guard case` pattern (no throwing call to wrap), so no changes there.

## Test plan
- [x] `xcodebuild build-for-testing` — 0 errors
- [x] `xcodebuild test -only-testing:VivaDictaTests` — 333 tests passed, 0 failed
- [x] Affected suites in `CloudTranscriptionTests` — all green
- [ ] Full CI test run

### DIFF
--- a/Modules/CloudTranscription/Tests/CloudTranscriptionTests/GroqTranscriptionServiceTests.swift
+++ b/Modules/CloudTranscription/Tests/CloudTranscriptionTests/GroqTranscriptionServiceTests.swift
@@ -217,15 +217,15 @@ struct GroqTranscriptionServiceTests {
         let audio = try makeAudioFile()
         let sut = makeService(networkService: networkService)
 
-        do {
+        let error = try await #require(throws: CloudTranscriptionError.self) {
             _ = try await sut.transcribe(audioURL: audio)
-            Issue.record("expected apiRequestFailed to throw")
-        } catch let CloudTranscriptionError.apiRequestFailed(statusCode, message) {
-            #expect(statusCode == 401)
-            #expect(message.contains("invalid key"))
-        } catch {
-            Issue.record("expected CloudTranscriptionError.apiRequestFailed, got \(error)")
         }
+        guard case let .apiRequestFailed(statusCode, message) = error else {
+            Issue.record("expected CloudTranscriptionError.apiRequestFailed, got \(error)")
+            return
+        }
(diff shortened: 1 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #298, opened 2026-05-26
Inject AIService/TranscriptionManager via init instead of fallback singletons

n0an:
## Summary

- `CloudModelConfigurationView` now receives `aiService` from `ModelsView` instead of `@State private var aiService = AIService()`. The old form constructed a fresh real keychain/network/OAuth stack on every view init, defeating the protocol-based test setup the rest of the codebase relies on.
- `RecordViewModel` captures `transcriptionManager` and `aiService` as stored properties at init time, removing the `?? TranscriptionManager()` / `?? AIService()` fallbacks on the computed accessors. Those fallbacks silently spun up real services whenever the weak `appState` was nil - including in tests. The new init takes optional overrides so tests can pass mocks directly.
- `AppState.init` now accepts `transcriptionManager`, `aiService`, and `presetManager` as defaulted parameters, so tests can substitute mocks without changing the production call site in `VivaDictaApp`.

These three changes unblock testing `RecordViewModel` and `AppState` against mock services - previously documented as blocked in `RecordViewModelTests.swift`.

## Test plan

- [x] `xcodebuild build` for iPhone 17 Pro Max / iOS 26.4 - succeeds (0 errors)
- [x] `xcodebuild test` - all 545 tests across all suites pass
- [ ] Smoke test in simulator: open Models screen, add an API key for a cloud model, verify save flow
- [ ] Smoke test: trigger a recording from the main view and from the keyboard extension

### DIFF
--- a/VivaDicta/AppState.swift
+++ b/VivaDicta/AppState.swift
@@ -120,20 +120,25 @@ class AppState {
     /// Audio file URL received via "Open With" from Files app.
     var openedAudioFileURL: URL?
 
-    init(modelContainer: ModelContainer) {
-        transcriptionManager = TranscriptionManager()
-        aiService = AIService()
-        presetManager = PresetManager(userDefaults: UserDefaultsStorage.shared)
-        aiService.presetManager = presetManager
-        PresetMigrationService.migrateIfNeeded(presetManager: presetManager, aiService: aiService)
+    init(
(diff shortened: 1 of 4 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #301, opened 2026-05-27
Backfill tests for APIKeyMigrationService and ReminderExtractionService

n0an:
## Summary

Wave A of the test coverage backfill. Both services were flagged in the audit as untested but already accepting their primary dependency via init - so they were nominally testable. In practice one needed a small DI tweak first.

**APIKeyMigrationService**
- Added \`sourceDefaults\` and \`flagDefaults\` init params (defaulting to \`UserDefaultsStorage.shared\` and \`UserDefaults.standard\`). The service used to read and write through both singletons directly, so any test would have polluted process-global state. Production behavior is unchanged - the \`.shared\` factory passes the real defaults.
- 8 new tests cover: flag short-circuit, single + multi-provider key migration, skip-empty-and-missing source values, custom transcription key migration, flag set after run (with and without keys to move), idempotency.

**ReminderExtractionService**
- 5 new tests pin \`canExtractReminders(using:)\` for the explicit reminder-extractor path:
  - false when a cloud provider requires a key that isn't connected
  - true for Ollama + model (no key needed)
  - false when extractor has no model
  - false for Copilot (hardcoded unsupported)
  - falls back to \`mode.aiProvider\` when no explicit extractor is set
- \`extractAndPersist\` makes real network/Apple FM calls and is out of scope here. \`AppleFoundationModelAvailability.isAvailable\` is environment-dependent on the simulator, so any test that would fall through to the Apple FM branch is intentionally skipped.

## Test plan

- [x] \`xcodebuild build\` for iPhone 17 Pro Max / iOS 26.4 - 0 errors
- [x] \`xcodebuild test\` - all green (557 tests; +13 from this PR)

### DIFF
--- a/VivaDicta/Services/APIKeyMigrationService.swift
+++ b/VivaDicta/Services/APIKeyMigrationService.swift
@@ -12,20 +12,32 @@ import os
 
 /// One-time migration of API keys from UserDefaults (App Group) to Keychain (iCloud sync).
 final class APIKeyMigrationService: Sendable {
-    static let shared = APIKeyMigrationService(keychain: DefaultKeychainService())
+    static let shared = APIKeyMigrationService(
+        keychain: DefaultKeychainService(),
+        sourceDefaults: UserDefaultsStorage.shared,
+        flagDefaults: UserDefaults.standard
+    )
 
     private let keychain: any KeychainService
+    private let sourceDefaults: UserDefaults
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #302, opened 2026-05-27
Backfill tests for DictionaryMigrationService

n0an:
## Summary

Same shape as #301's APIKeyMigrationService work. The service used to read source data from \`UserDefaultsStorage.appPrivate\` and the completion flag from \`UserDefaults.standard\` directly, which made the migration logic untestable without polluting process-global state.

- Adds \`sourceDefaults\` and \`flagDefaults\` init params (defaulting to the same singletons). \`.shared\` factory passes the real defaults; production behavior is unchanged.
- Init is no longer \`private\` so tests can construct the service with isolated UserDefaults.
- 7 new tests:
  - flag short-circuit when already migrated
  - vocabulary words migrate into SwiftData
  - empty/whitespace vocabulary entries are skipped
  - word replacements migrate from the legacy JSON payload
  - completion flag is set after a run
  - completion flag is set even when there's nothing to migrate
  - idempotency: second call doesn't duplicate rows

## Test plan

- [x] \`xcodebuild build\` for iPhone 17 Pro Max / iOS 26.4 - 0 errors
- [x] \`xcodebuild test\` - all green (+7 new tests)

### DIFF
--- a/VivaDicta/Services/DictionaryMigrationService.swift
+++ b/VivaDicta/Services/DictionaryMigrationService.swift
@@ -4,17 +4,29 @@ import SwiftData
 import os
 
 class DictionaryMigrationService {
-    static let shared = DictionaryMigrationService()
+    static let shared = DictionaryMigrationService(
+        sourceDefaults: UserDefaultsStorage.appPrivate,
+        flagDefaults: UserDefaults.standard
+    )
+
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #303, opened 2026-05-27
Memoize derived state in TranscriptionsContentView

n0an:
## Summary

Three small fixes that cut O(n) work on the full transcription array out of the hot path of `TranscriptionsContentView`. No behavior changes - just less work per body re-eval and per CloudKit sync.

### What changed

- **`availableSourceTags` is now `@State`**, not a computed property. Previously it rescanned the entire `allTranscriptions` array on every body re-eval (selection toggle, search keystroke, mode switch, etc.). Now it's recomputed only when the underlying data changes.
- **Added `transcriptionsByID: [UUID: Transcription]`** for O(1) lookups. Replaces two O(n) `first(where: { $0.id == ... })` call sites:
  - `transcription(for:)` (called per smart-search hit in `smartDisplayedTranscriptions`)
  - The smart-search ordering pass inside `performDebouncedSearch`
- **Short-circuited the double filter scan** inside `onChange(of: allTranscriptions)`. The two `filterTranscriptionsByActiveTags(...)` calls used to fire on every CloudKit sync, even when no tag filter was active. Now they only run when `hasActiveTagFilter` is true (the only case where the result actually affects the outcome).

### Why

For users with hundreds of transcriptions, the previous code did O(n) work on every SwiftUI body invocation (recomputing source tags) and on every CloudKit-triggered `@Query` refresh (scanning the array twice for filter-reset accounting + scanning again for smart-search ordering). None of this was visible as a single slow frame, but it adds up under heavy scroll / search / sync.

### Why not pagination

Considered and rejected. Pagination would force a parallel non-paginated path for tag filtering and "what source tags exist" (since those need the full set), and the real cost on this view is not memory or row rendering (SwiftUI `List` is already lazy) - it's the O(n) recomputation pattern fixed here.

## Test plan

- [x] `xcodebuild` succeeds with 0 errors
- [ ] Launch app, scroll the transcriptions list, confirm no behavior change
- [ ] Search (keyword and smart) returns the same results as before
- [ ] Tag filter bar still shows the correct available source tags
- [ ] Delete a transcription with a tag filter active - tag filter resets correctly when the last matching note is deleted
- [ ] New transcription appears at the top with the same insertion animation

### DIFF
--- a/VivaDicta/Views/TranscriptionsContentView.swift
+++ b/VivaDicta/Views/TranscriptionsContentView.swift
@@ -55,6 +55,8 @@ struct TranscriptionsContentView: View {
     @State private var selectedSourceTags: Set<String>
     @State private var selectedUserTagIds: Set<UUID>
     @State private var searchMode: TranscriptionSearchMode = .all
+    @State private var availableSourceTags: [String] = []
+    @State private var transcriptionsByID: [UUID: Transcription] = [:]
 
     private let topAnchorID = "topAnchor"
     private let logger = Logger(category: .transcriptionsContentView)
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
@claude please re-review - addressed the transcriptionsByID staleness concern

n0an:
@codex please re-review - addressed the transcriptionsByID staleness concern

PR #306, opened 2026-05-29
Fix tag filter sync + inherit filter tags on new recordings

n0an:
## Summary

Two related tag/filter improvements.

### 1. Inherit the active tag filter when starting a recording
When a tag filter is active on the main screen, starting a new recording now pre-selects those filter tags so the saved note stays visible under the current filter instead of disappearing (and forcing a filter reset).

- `startCaptureAudio` gains an `initialTagIds: Set<UUID> = []` parameter; the stale-clear line becomes `pendingTagIds = initialTagIds` (clears *and* seeds in one step). Default empty set leaves keyboard/intent call sites untouched.
- `MainView.startRecording()` seeds `savedNotesFilter.userTagIds`, scoped to `.newNote` destinations (append-to-note recordings are left untouched).
- Seeded tags appear pre-selected in the recording sheet (count badge + highlighted chips) and stay reversible before saving.
- Only user tags are seeded - source tags (`app`, `keyboard`, ...) are system-assigned, and app recordings already auto-match a source-tag filter.

### 2. Fix stale tag filter when toggling chips on the main screen
The tag filter had **two sources of truth**:
- `MainView.savedNotesFilter` - drives the toolbar filter button, the `NotesFilterView` ("Disable Filter") screen, and persistence.
- `TranscriptionsContentView`'s local `selectedSourceTags`/`selectedUserTagIds` - drives the on-screen chips and the actual list filtering.

Sync was **one-way** (`savedFilter` → local), so tapping a chip updated the list but never wrote back - leaving the toolbar and filter screen showing a stale "active" filter.

Fix: make `savedFilter` a `@Binding` (single source of truth) and add guarded two-way sync:
- `syncFilterUpstream()` propagates chip taps back into `savedNotesFilter`.
- The downstream `onChange(of: savedFilter)` only assigns when values differ, so the two directions converge in one pass instead of bouncing.

Result: deselecting the last chip clears the filter (toolbar inactive, screen "not applied"); selecting a chip from the bar activates and persists it. The delete-triggered `shouldResetTagFilter` auto-reset now also clears the persisted filter, which previously had the same staleness bug.

## Testing

- `xcodebuild` build succeeds for the iOS Simulator target.
- Manual verification of chip toggling, filter-screen consistency, and recording-while-filtered recommended on-device.

### DIFF
--- a/VivaDicta/Models/TranscriptionModelProvider.swift
+++ b/VivaDicta/Models/TranscriptionModelProvider.swift
@@ -290,6 +290,19 @@ enum TranscriptionModelProvider: String, Sendable, Codable, CaseIterable, Identi
                 supportManyLanguages: true,
                 supportedLanguages: cohereLanguages
             ),
+            
+            CloudModel(
+                name: "grok-stt",
+                displayName: "xAI Speech-to-Text",
+                description: "xAI's hosted speech-to-text endpoint with natural formatting and multilingual support. Uses the same xAI API key as Grok chat. No auto-detect - pick a language.",
+                provider: .xai,
+                recommended: true,
+                speed: 0.9,
+                accuracy: 0.95,
+                cost: 0.5,
+                supportManyLanguages: true,
+                supportedLanguages: xaiLanguages
+            ),
 
             CloudModel(
                 name: "ink-whisper",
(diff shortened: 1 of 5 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #308, opened 2026-05-29
Overwrite existing file on folder re-export instead of duplicating

n0an:
## Summary

Exporting a note to the chosen folder a second time created a new file with a `-2` suffix instead of updating the existing one. Since the export filename is deterministic per note (`VivaDicta-YYYY-MM-DD_HHMMSS.md`, derived from the note's timestamp), the same note always maps to the same filename - so re-exporting should refresh that single file, not pile up copies.

- Removed the `uniqueDestination` collision-uniquing (which appended `-2`, `-3`, …) in `FolderExportService`.
- Now writes directly to the deterministic path; the existing atomic write replaces any existing file.
- Applies to both the manual "Export to Folder" button and auto-export, so re-transcribing/re-exporting a note keeps one file per note.

## Testing

- `xcodebuild` Debug build for iPhone 17 Pro Max (iOS 26.4): **BUILD SUCCEEDED**.

## Notes / tradeoff

Two *different* notes created in the exact same second would now map to the same filename and overwrite each other. For second-granularity voice-note timestamps this is effectively impossible in normal use. If we want a hard guarantee, we could fold a short note-ID into the filename - out of scope here.

### DIFF
--- a/VivaDicta/Services/FolderExportService.swift
+++ b/VivaDicta/Services/FolderExportService.swift
@@ -147,7 +147,10 @@ enum FolderExportService {
         }
         defer { folderURL.stopAccessingSecurityScopedResource() }
 
-        let destination = uniqueDestination(in: folderURL, baseFilename: item.filename)
+        // The filename is deterministic per note (derived from its timestamp),
+        // so re-exporting the same note overwrites its previous file rather than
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #310, opened 2026-05-29
Add Default AI Mode setting for AI Actions in note detail

n0an:
## Summary

Adds a **Default AI Mode** picker so users can apply AI presets in a note even when the currently selected mode has no AI processing configured.

Previously, opening the **AI Actions** sheet on a note whose active mode lacked an AI provider showed the *"AI Processing Not Configured"* empty state. The user had to manually open the sheet's top-left mode picker and switch to a mode with AI configured before any preset would apply.

Now, a single **Default AI Mode** picker lives in **Advanced → Note Detail** (next to *Append with Voice*). When set, and the active mode can't do AI, the AI Actions sheet **pre-selects that default mode** and jumps straight to the preset list. The top-left mode picker still lets the user switch (including back to the unconfigured mode).

## Changes

- **`UserDefaultsStorage.swift`** - new app-private key `defaultAIModeId` (stores the mode's UUID string; empty = none).
- **`AdvancedSettingsView.swift`** - new `Default AI Mode` menu picker in the existing *Note Detail* section:
  - Options: **None** + every mode that is AI-configured (`isProperlyConfigured(for:requirePreset: false)` - the same filter the AI Actions sheet already uses for mode overrides).
  - A sanitizing binding snaps the selection back to **None** if the saved mode is later deleted or has AI disabled, so the picker never shows a stale/blank value.
- **`TranscriptionDetailView.swift`**:
  - Reads `defaultAIModeId` and passes it into `PresetPickerSheet` as `preselectedOverrideModeId`.
  - `PresetPickerSheet` gains an explicit `init` that seeds its existing `overrideModeId` with the default **only when** the active mode isn't AI-configured **and** the default mode is still present in `availableModes`. Seeding in `init` (via `State(initialValue:)`) avoids a one-frame flash of the empty state.

No data-model changes - this reuses the sheet's existing mode-override mechanism, so downstream variation generation already routes through the chosen mode.

## Behavior

- Active mode **has** AI → unchanged; uses the active mode.
- Active mode **lacks** AI + default set → sheet opens to presets with the default mode pre-selected.
- Default mode deleted / AI turned off → falls back to today's empty state.

## Testing

- `xcodebuild` Debug build for iOS Simulator (iPhone 17 Pro Max, OS 26.4): **BUILD SUCCEEDED**.
- Manual reasoning over the edge cases above (stale default, active-mode-configured, no configured modes).

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
+++ b/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
@@ -93,6 +93,10 @@ public enum UserDefaultsStorage {
         public static let isLiveTranslationEnabled = "isLiveTranslationEnabled"
         public static let isStripTrailingPeriodEnabled = "isStripTrailingPeriodEnabled"
 
+        /// UUID string of the mode used for AI actions in note detail when the
+        /// currently selected mode has no AI processing configured. Empty = none.
+        public static let defaultAIModeId = "defaultAIModeId"
+
         // Notes filter
         public static let savedNotesFilterSourceTags = "savedNotesFilterSourceTags"
         public static let savedNotesFilterUserTagIds = "savedNotesFilterUserTagIds"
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #311, opened 2026-05-30
Fix recording-orb freeze and add orb visualization picker

n0an:
## Summary

Fixes a recording-screen freeze caused by the ASCII recording orb overloading the CPU, reworks the orb setting into a three-way picker defaulting to the lighter visualization, and applies a Form-picker styling convention.

### Recording-orb freeze fix
A user reported the app freezing roughly 6 seconds into every recording with on-device models on an iPhone 15 Pro Max (iOS 26.5). Root cause: the ASCII orb's `Canvas` was far too expensive per frame and used `rendersAsynchronously: true`, which let `TimelineView(.animation)` queue draw closures faster than they could complete - an unbounded backlog that progressively collapsed the frame rate and froze the app once thermal headroom was reduced (charging + first-launch on-device model loads).

`ASCIIOrbView` changes:
- Synchronous `Canvas` (removes the async frame backlog - the actual runaway mechanism)
- Cap the render cadence at 20 fps (a 120 Hz ProMotion display no longer renders up to 6x as often)
- Grid reduced from 40×40 to 32×32
- Drop the high-frequency noise octave (3 → 2 noise evaluations per cell)

Net: ~90% less per-second work on a 120 Hz device, and the runaway backlog is gone.

### Recording Orb picker
- Replaces the "ASCII Recording Orb" on/off toggle with a "Recording Orb" picker: **Particles** (default), **ASCII**, **None**
- Defaults everyone to the lighter Particles orb; **None** hides the visualization entirely (and shrinks the recording sheet)
- New `recordingOrbStyle` preference, no migration - existing installs simply fall through to the Particles default

### Form-picker convention
- Removes redundant `.pickerStyle(.menu)` / `.tint(.primary)` from `Form` pickers so the entire row stays tappable, and documents the rule in `AGENTS.md`

## Testing
- Debug build passes (`xcodebuild`, iPhone 17 Pro Max simulator, OS 26.4)
- Root cause confirmed externally: the reporting user verified that disabling the ASCII orb eliminated the freeze; these changes remove the orb's runaway cost so it no longer needs disabling
- Picker wiring verified to compile; not yet visually QA'd on a 120 Hz device under load

## Notes
- No data migration. The legacy `isASCIIOrbEnabled` key is removed and unused.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -167,6 +167,7 @@ Run `./build-docc.sh` from the project root on the `main` branch. The script bui
 - Never use `ObservableObject`; always prefer `@Observable` classes instead.
 - Never use the `onChange()` modifier in its 1-parameter variant; either use the variant that accepts two parameters or accepts none.
 - Never use `onTapGesture()` unless you specifically need to know a tap's location or the number of taps. All other usages should use `Button`.
+- Don't add `.pickerStyle(.menu)` or `.tint(.primary)` to a `Picker` inside a `Form` (or `List`). A `Form` `Picker` already defaults to a menu where the entire row is tappable; adding those modifiers is redundant and makes only part of the row tappable. Leave them off so the whole row stays tappable.
 - Never use `Task.sleep(nanoseconds:)`; always use `Task.sleep(for:)` instead.
 - Never use `UIScreen.main.bounds` to read the size of the available space.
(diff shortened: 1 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #313, opened 2026-06-01
Add Ollama Cloud AI provider for text processing

n0an:
## Summary

Adds **Ollama Cloud** as a new cloud-based AI text provider, distinct from the existing local **Ollama** provider. Both reuse the same Ollama icon asset.

Ollama Cloud authenticates with an API key (Bearer token, obtained from https://ollama.com/settings/keys) against the OpenAI-compatible endpoint at `https://ollama.com/v1/chat/completions`.

### Why this approach

Ollama Cloud is modeled as a standard OpenAI-compatible cloud provider. As a result, request handling, response streaming, multi-turn chat, reminder extraction, and API-key verification all flow through the existing `default` code paths - so **no changes were needed in the request layer or the UI**. The provider list, API-key entry screen, and mode editor are already generic over `AIProvider`.

The cloud model catalog changes over time, so models are fetched dynamically from the `/v1/models` endpoint (which requires the API key) after a key is verified and on launch when connected, mirroring the OpenRouter/HuggingFace pattern. A curated static fallback list keeps the model picker populated before the first fetch.

## Changes

**`AIProvider.swift`** - new `.ollamaCloud` case wired into every switch and array:
- `displayName` -> "Ollama Cloud"; `iconName` -> "ollama" (shared asset)
- `baseURL` -> `https://ollama.com/v1/chat/completions`; `defaultModel` -> `gpt-oss:120b`
- `keychainKey` -> `ollamaCloudAPIKey`; `apiKeyURL` -> settings/keys page
- Added to `supportsResponseStreaming`, `cloudProviders`, `generalProviders`
- `requiresAPIKey` returns `true` automatically (not in the exclusion list)
- `availableModels` -> curated fallback list of known cloud models

**`AIService.swift`**:
- New `ollamaCloudModels` observable property + load/save helpers + UserDefaults persistence
- `fetchOllamaCloudModels()` - fetches `/v1/models` with the Bearer key (the cloud listing requires auth, unlike the other catalogs)
- Fetches after key verification and on launch when connected
- `getAvailableModels(for:)` returns the live list, falling back to the static one

**`UserDefaultsStorage.swift`** - new `ollamaCloudModels` key.

## Behavior

Settings -> AI Providers -> Cloud now shows both "Ollama" (local, server config) and "Ollama Cloud" (Add API Key) adjacent, both with the Ollama logo. Adding a key verifies it against the cloud endpoint, then populates the live model list for the mode editor's model picker.

## Testing

- `xcodebuild` build succeeds with no new errors or warnings.
- Manual verification on device/simulator with a real Ollama Cloud API key is still recommended to confirm the live `/v1/models` response shape and end-to-end enhancement.

## Notes

- The static fallback model names are a best-effort snapshot of the current cloud catalog and may drift over time, but the dynamic fetch replaces them with the live list once a valid key is entered. `gpt-oss:120b` (the verification/default model) is the documented flagship cloud model.
- The local Ollama provider is unchanged.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
+++ b/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
@@ -63,6 +63,7 @@ public enum UserDefaultsStorage {
         public static let huggingFaceModels = "huggingFaceModels"
         public static let ollamaModels = "ollamaModels"
         public static let ollamaServerURL = "ollamaServerURL"
+        public static let ollamaCloudModels = "ollamaCloudModels"
 
         // iCloud
         public static let isICloudSyncEnabled = "isICloudSyncEnabled"
--- a/VivaDicta/Services/AIEnhance/AIProvider.swift
+++ b/VivaDicta/Services/AIEnhance/AIProvider.swift
@@ -234,6 +234,14 @@ enum AIProvider: String, CaseIterable, Identifiable, Codable {
         .ollamaCloud,
         .customOpenAI]
 
+    /// Cloud providers eligible to act as a reminder extractor.
+    ///
+    /// Excludes Ollama Cloud: reminder extraction relies on structured
+    /// (`json_schema`) outputs, which Ollama Cloud does not support
+    /// (https://docs.ollama.com/capabilities/structured-outputs). Local Ollama
+    /// does support them, so it stays eligible.
+    static let reminderExtractorCloudProviders: [AIProvider] = cloudProviders.filter { $0 != .ollamaCloud }
+
     /// Local AI providers that run on-device or local network (no API key needed)
     static let localProviders: [AIProvider] = [
         .apple,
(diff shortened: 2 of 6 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Addressed review feedback in d3d666e3b:

- **Codex P2 (reminder extraction):** Ollama Cloud lacks structured-output (`json_schema`) support, so reminder extraction would reliably fail while showing as ready. `CloudReminderExtractionProvider.canExtract` now returns `false` for it (the runtime gate for both auto-extraction and the manual Extract Tasks action), and it's no longer offered as a reminder provider - a new `AIProvider.reminderExtractorCloudProviders` list drives the picker and auto-select, and the prefill path no longer copies a non-eligible AI provider into the reminder extractor. (The local Ollama provider, which does support structured outputs, stays eligible.)
- **Claude #1 & #2 (`fetchOllamaCloudModels`):** now skips on an empty API key and logs the no-key case at debug level instead of error.

Left the remaining nits (shared icon is intentional per the requirement; sort order matches peer providers; type filter is speculative). Build passes.

PR #314, opened 2026-06-01
Add AssemblyAI cloud transcription provider

n0an:
## Summary

Adds **AssemblyAI** as a cloud speech-to-text (transcription) provider. It uses AssemblyAI's async **upload → create → poll** flow:

1. Upload the raw audio bytes to `POST /v2/upload` to obtain a hosted `upload_url`.
2. Create a transcript job at `POST /v2/transcript` with the `audio_url` and options.
3. Poll `GET /v2/transcript/{id}` until status is `completed`, then read the text (and `utterances` when diarization is on).

Authentication uses the **raw API key in the `Authorization` header** (AssemblyAI does not use a `Bearer` prefix). The request supports automatic language detection (`language_detection`) or a chosen `language_code`, speaker diarization (`speaker_labels`), and custom vocabulary (`word_boost`).

Two models are exposed in the picker: **Universal-3 Pro** (recommended) and **Universal-2**.

## Why this shape

AssemblyAI is modeled on the existing async upload+poll providers (Gladia/Speechmatics) and threaded through the standard transcription layers, so it reuses all existing infrastructure:

- `AssemblyAITranscriptionService` (new) conforms to `TranscriptionService`
- `TranscriptionProvider` enum + `DefaultTranscriptionEngine` service dispatch
- `TranscriptionManager.makeProvider` builds the config (model, language, vocabulary, diarization)
- `TranscriptionModelProvider` gets the picker metadata, ordering (grouped with the other async providers), default model, `mappedAIProvider` mapping, and two `CloudModel` catalog entries
- `AIProvider` gets the case + display name, icon (`assemblyai-color`), API-key URL, base URL, default model, keychain key (`assemblyaiAPIKey`)
- `AIService.verifyAssemblyAIAPIKey` does a cheap auth probe via `GET /v2/transcript?limit=1`

Like the other speech-only providers, AssemblyAI is intentionally kept out of the AI-text provider arrays. The transcription provider/model picker, API-key entry, and verification are all data-driven through `mappedAIProvider`, so no UI changes were needed.

## Testing

- `xcodebuild` build succeeds with no new errors or warnings (SwiftPM modules + app target).
- Not yet exercised against a live AssemblyAI key - an end-to-end run with a real key is recommended to confirm the request/response shape.

## Notes / caveats

- Follows the linked docs which use `speech_models` (array, no default); sends the user's selected model. If an account is on the older API expecting singular `speech_model`, the create call would need adjusting.
- `word_boost` (custom vocabulary) is sent guarded; the Universal models may ignore it, but it won't break core transcription.
- Languages use the full shared language set (auto → `language_detection`, otherwise `language_code`), matching the other broad providers; a rare code could be rejected by AssemblyAI.
- AssemblyAI's transcript endpoint has no simple target-language translation parameter, so inline translation is intentionally not offered (same as Deepgram/Cohere/xAI).


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Addressed review feedback in 01ce20c23:

- **Critical (live test hit a 400) + Claude #7:** `word_boost` is rejected by Universal-3. Switched to `keyterms_prompt`, which both Universal-2 and Universal-3 Pro support; capped terms at 200 (Universal-2's limit).
- **Codex P1:** the default `universal-3-pro` now sends `speech_models: ["universal-3-pro", "universal-2"]` so languages outside U3 Pro's six (en/es/de/fr/pt/it) fall back to Universal-2 (99 languages).
- **Claude #1:** job-level failures now throw with `statusCode: -1` so the real error message is shown instead of a misleading "service unavailable" notice.
- **Claude #2:** the completed-response decode now uses bare `try` so schema drift surfaces instead of being masked as "no transcription".
- **Claude #5:** the model-name fallback is folded into the new `speech_models` logic.
- **Claude #7 (confirmed live):** `speech_models` (plural array) is accepted on the test account.

Deferred (nits): poll interval (3s per AssemblyAI's recommended cadence), nil-speaker fold (matches Speechmatics), and unit tests (consistent with the other async providers, which have none). Build passes.

PR #315, opened 2026-06-04
Add speaker diarization to ElevenLabs and xAI transcription

n0an:
## Summary

ElevenLabs Scribe and xAI grok-stt both support speaker diarization, but their transcription services only ever returned plain text - so the global **Speaker Labels** toggle silently did nothing for those two providers. This wires both into the existing diarization path, closing the gap.

After this change, every transcription provider whose API supports diarization implements it: AssemblyAI, Deepgram, Mistral, Gladia, Soniox, Speechmatics, WhisperKit (local), **ElevenLabs**, and **xAI**.

## Changes

- **ElevenLabs** (`ElevenLabsTranscriptionService`): sends `diarize=true` when enabled, parses the word-level `speaker_id`, and groups consecutive words per speaker. ElevenLabs interleaves `spacing` tokens, so word `text` is concatenated directly to preserve boundaries.
- **xAI** (`XaiTranscriptionService`): sends `diarize=true` (before the mandatory-last `file` field), parses the 0-based integer `speaker` per word. xAI emits no spacing tokens, so words in a turn are re-joined with single spaces. Confirmed against xAI's REST reference that the batch `/v1/stt` endpoint has no `model` field, so the existing request shape is unchanged.
- Both reuse the shared `SpeakerDiarizationFormatter` and return `.speakerAttributed(...)`, matching the 7 providers that already do.
- `TranscriptionManager.makeProvider` passes the global `diarizationEnabled` flag through to both providers.
- `SettingsView`: the Speaker Labels toggle description now lists ElevenLabs and xAI.

## Testing

- `swift test` on the CloudTranscription package: **24 tests pass** across both suites.
- Added 4 diarization cases to `XaiTranscriptionServiceTests` (field omitted by default, sent when enabled, word grouping, disabled-stays-plain).
- Added a new `ElevenLabsTranscriptionServiceTests` suite (none existed) covering success path, request shape, and the same diarization scenarios.
- Full app target builds: **BUILD SUCCEEDED** (iOS Simulator, iPhone 17 Pro Max, OS 26.4).

## Notes / out of scope

- **Parakeet (local/FluidAudio)**: FluidAudio ships a diarizer, but that's a separate local-pipeline integration (model wiring + download), not a one-field cloud change. Candidate for a follow-up to reach local parity with WhisperKit.
- **OpenAI**: current models (`gpt-4o-transcribe`, `whisper-1`) genuinely don't diarize. Unlocking it means adding `gpt-4o-transcribe-diarize` as a selectable model - a product decision, not a gap.
- **Gemini**: prompt-only diarization, no structured API - left as plain.
- **Cartesia, Cohere, Groq, Custom**: confirmed no diarization support; correctly plain.

No breaking changes.

### DIFF
--- a/Modules/CloudTranscription/Sources/CloudTranscription/ElevenLabsTranscriptionService.swift
+++ b/Modules/CloudTranscription/Sources/CloudTranscription/ElevenLabsTranscriptionService.swift
@@ -16,11 +16,18 @@ public struct ElevenLabsTranscriptionService: TranscriptionService, Sendable {
         public let modelName: String
         /// BCP-47 code, or `"auto"` to let ElevenLabs detect.
         public let language: String
+        public let isSpeakerDiarizationEnabled: Bool
 
-        public init(apiKey: String, modelName: String, language: String = "auto") {
+        public init(
+            apiKey: String,
+            modelName: String,
+            language: String = "auto",
+            isSpeakerDiarizationEnabled: Bool = false
+        ) {
             self.apiKey = apiKey
             self.modelName = modelName
             self.language = language
+            self.isSpeakerDiarizationEnabled = isSpeakerDiarizationEnabled
         }
     }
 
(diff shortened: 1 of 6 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #317, opened 2026-06-05
Replace keyboard swipe-back toast with full-screen prompt

n0an:
## Summary

The small top toast that told users to **swipe back to continue keyboard dictation** was too easy to miss. Some users never realized they needed to return to the host app, so dictation appeared to do nothing.

This replaces the toast with a **full-screen cover** that's hard to miss.

### What the new prompt shows
- 🟢 **"Keyboard active"** status pill
- A prominent **"Swipe back to continue"** headline (brand orange)
- The **VivaDicta app icon**, centered
- A short explanation of *why* the app switch is required (Apple requires it to activate the mic)
- A **Dismiss** button

### Behavior
- **Three dismissal paths:** swipe-back (app resigns active, handled by the existing scene-phase hook), the Dismiss button, or an automatic **30-second** timeout.
- The recording sheet is **suppressed while the prompt is shown** (SwiftUI presents only one modal at a time) and **restored** if recording is still running when the prompt closes.
- Trigger paths are unchanged - the prompt fires in the same keyboard-return-failure cases the toast did (no host URL scheme, missing `hostId`, no model selected, prewarm failure).

### Code changes
- **New:** `KeyboardReturnPromptView.swift` (full-screen cover).
- **Removed:** `KeyboardFlowToast.swift`.
- **`MainView.swift`:** dropped the top-overlay toast; added `.fullScreenCover` as a sibling of the recording sheet; added guard/restore logic so the two modals never collide.
- **`AppState.swift` + `VivaDictaApp.swift`:** renamed `showKeyboardFlowToast` → `showKeyboardReturnPrompt` to reflect the new presentation (all 11 call sites updated).

## Testing
- `xcodebuild` build succeeds for the `VivaDicta` scheme (iPhone 17 Pro Max, iOS 26.4).
- The live trigger path (keyboard mic → host app with no URL scheme / missing `hostId`) is hard to reproduce in the simulator; recommend an on-device keyboard test before release.

## Notes
- No breaking changes. Pure UX/presentation change plus an internal property rename.


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #318, opened 2026-06-05
Live Translation copy buttons + filler fix for translated transcriptions

n0an:
Two related-area changes from one session, kept as separate commits for easy review.

## 1. Copy buttons in Live Translation (`9613f9a14`)

Each transcript column now shows a **Copy** button at its bottom once it has text:
- Left-aligned under the **source** column, right-aligned under the **translation** column.
- Copies that column's plain text via `ClipboardManager`, fires a success haptic, and briefly flips to a green **"Copied"** checkmark - the same idiom used in the transcription list (`TranscriptionRowView` / `AnimatedCopyButton`).
- Only appears when the column has content (mirrors how "Save as note" only shows with text).

## 2. Fix filler stripping for translated transcriptions (`d8b43730b`)

**Bug:** With inline translation (e.g. Soniox **Spanish → Russian**), the transcription output is in the *target* language, but `TranscriptionManager` passed the *source* language to `TranscriptionOutputFilter`. So Russian fillers (`э-э`, `ээ`, `эм`) were never stripped - output looked like:

> Э-э, привет, э-э, сейчас я говорю по-испански, но, э-э, давайте посмотрим...

**Fix:** Resolve the output language to the translation target when set, falling back to the source transcription language, and pass that to the filter. Added a `TranscriptionManager.outputLanguage(...)` helper plus regression tests (resolution logic + the Spanish→Russian filler-stripping case).

**Live Translation is unaffected** - its streaming screen renders raw Soniox tokens directly and never runs through `TranscriptionOutputFilter`, so streaming output keeps its fillers (intended, since it's a live stream).

## Testing
- App builds clean (`xcodebuild`, iPhone 17 Pro Max, iOS 26.4).
- New tests added in `TranscriptionOutputLanguageTests`. Note: the full `xcodebuild test` action currently can't run because of a **pre-existing, unrelated** Swift 6 actor-isolation compile error in the `VivaDictaUITests` target (not touched by this PR). The new logic is also covered transitively by the existing, passing `filter(..., language: "ru")` tests.

## Notes
- No breaking changes.

### DIFF
--- a/VivaDicta/Services/Transcription/TranscriptionManager.swift
+++ b/VivaDicta/Services/Transcription/TranscriptionManager.swift
@@ -160,9 +160,16 @@ class TranscriptionManager {
             progress: progressHandler
         )
 
+        // When inline translation is active (e.g. Soniox Spanish → Russian) the
+        // output text is in the TARGET language, so the filler set must match the
+        // target, not the source transcription language - otherwise target-language
+        // fillers (e.g. Russian "э-э") survive.
         var result = TranscriptionOutputFilter.filter(
             transcriptionResult.text,
-            language: currentMode.transcriptionLanguage
+            language: Self.outputLanguage(
+                transcriptionLanguage: currentMode.transcriptionLanguage,
+                translationTargetLanguage: currentMode.translationTargetLanguage
+            )
         )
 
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #325, opened 2026-06-17
Move EnhancementError into AICore

n0an:
## Phase 3: move EnhancementError into AICore

Relocates the `EnhancementError` `LocalizedError` enum from `AIService.swift` into the `AICore` module as a `public`, `Sendable` type. This is a small foundational move that **unblocks** moving the provider services into `AIProviders` - they all throw `EnhancementError`, so it has to live in the shared module first.

### Why this, and not the protocol (a plan reorder)

The original plan had Phase 3 define an `AITextProvider` protocol and conform the provider services in place. On inspection the services' `enhance`/`enhanceStreaming` signatures diverge too much for a clean "conform in place" (Anthropic takes `apiKey`/`model`; OpenAICompatible takes `url`/`headers`/`timeout`; Apple FM takes `samplingProfile`). So the providers move into `AIProviders` **first**, and the protocol is defined later once they're co-located and can be refactored to config-bound `init`. This PR is the foundational error-type move that the provider moves depend on.

### Changes

- `EnhancementError` moves into `AICore` (`public`, `Sendable`). It was a file-scope enum in `AIService.swift`, app-only, so - unlike the catalog enum - there is **no** extension `membershipExceptions` involvement.
- Behavior tests added: custom-message passthrough (`errorDescription` / `failureReason`) and structural completeness across every case. Copy is intentionally not over-asserted, since user-facing error strings may legitimately change.
- `import AICore` added to the 4 remaining consumers; the rest already import it from the catalog-enum move.

### Behavior

Pure relocation - **no behavior change.** Every case and message string is unchanged.

### Verification

- `AICoreTests` green (`swift test`).
- App + all three extensions + test bundle build green (`build-for-testing`, iOS 26.4 simulator) on the first pass.

### Next phase

Begin moving provider services into `AIProviders`, starting with the standalone `AnthropicService` (`AIProviders` gains a `Networking` dependency; `AIService` imports `AIProviders`).

### DIFF
--- a/Modules/AICore/Sources/AICore/EnhancementError.swift
+++ b/Modules/AICore/Sources/AICore/EnhancementError.swift
@@ -1,0 +1,50 @@
+import Foundation
+
+/// Errors that can occur during AI text enhancement.
+public enum EnhancementError: LocalizedError, Sendable {
+    case notConfigured
+    case invalidResponse
+    case enhancementFailed
+    case networkError
+    case serverError
+    case rateLimitExceeded
+    case customError(String)
+
+    public var errorDescription: String? {
+        switch self {
+        case .notConfigured:
+            return "AI provider not configured"
+        case .invalidResponse:
+            return "Invalid response from AI"
+        case .enhancementFailed:
+            return "AI processing failed"
+        case .networkError:
+            return "Network connection failed"
+        case .serverError:
+            return "Server error occurred"
+        case .rateLimitExceeded:
+            return "Rate limit exceeded"
+        case .customError(let message):
+            return message
+        }
+    }
+
+    public var failureReason: String {
+        switch self {
+        case .notConfigured:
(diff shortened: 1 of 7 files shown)


n0an:
@claude please review this PR. It moves the `EnhancementError` LocalizedError enum from AIService.swift into the AICore module (public + Sendable) - a pure relocation that unblocks moving the provider services into AIProviders. Focus on: behavior preservation (cases + message strings unchanged), public/Sendable correctness, and import coverage (any consumer missing the import, or any unnecessary one).

n0an:
@codex please review this PR. It moves the `EnhancementError` LocalizedError enum from AIService.swift into the AICore module (public + Sendable), should be a pure relocation. Focus on: any behavioral drift in the error cases or messages, public/Sendable correctness, and whether every throw/catch site still resolves the type (import coverage), including member-access sites like `error.failureReason` that don't name the type.

PR #327, opened 2026-06-17
Move AnthropicService into AIProviders

n0an:
## Phase 5: move AnthropicService into AIProviders

Relocates `AnthropicService` - the standalone Anthropic Messages API client (`POST /v1/messages`, non-streaming + SSE streaming + key verification) - from the app target into the `AIProviders` module as a `public` struct. **This is the first concrete provider to land in the impl module**, making `AIProviders` a real module rather than scaffolding.

### Changes

- **`AnthropicService` → `AIProviders`**, made `public` (struct + init + `enhance` / `enhanceStreaming` / `verifyAPIKey`). Only access-control modifiers were added - the HTTP headers (`x-api-key`, `anthropic-version`), status-code mapping (429/5xx/other), JSON body shape, and `content_block_delta` SSE parsing are byte-for-byte unchanged.
- **`AIProviders` gains a `Networking` dependency** (for `NetworkService`) and an internal `Logger+AIProviders.swift` extension, following the existing per-module logging pattern (`Logger+CloudTranscription`, `Logger+OAuth`) - methods are `internal` so they don't collide with the app's identically-named `Logger` extension.
- **`AnthropicServiceTests` (12 tests) → `AIProvidersTests`**, with its own `TestTags.swift` (`.networking`). `@testable import VivaDicta` becomes `import AIProviders`; test deps `AICore` / `Networking` / `NetworkingMocks` are wired.
- The empty `AIProviders` module/test placeholder files are removed now that the module has real content.
- `AIService` constructs the provider via `import AIProviders`; its three `AnthropicService(networkService:logger:baseTimeout:)` call sites are unchanged.

### Behavior

Pure relocation - **no behavior change.** The diff against the old file is exactly five `public` additions; a byte-level and whitespace-insensitive comparison confirms no other line changed.

### Verification

- 12 `AIProviders` tests pass (`swift test`); `AIProviders` also builds in isolation (`swift build`), proving no app symbol is required.
- App + all three extensions + test bundle build green (`build-for-testing`, iOS 26.4 simulator). No `project.pbxproj` change (synchronized-folder move).

### Next phase

Move `AppleFoundationModelService` (the other standalone provider, on-device) into `AIProviders`, then the `OpenAICompatibleService` cluster (with its `Ollama`/`CustomOpenAI` delegators).

### DIFF
--- a/Modules/AIProviders/Package.swift
+++ b/Modules/AIProviders/Package.swift
@@ -13,17 +13,24 @@ let package = Package(
     ],
     dependencies: [
         .package(path: "../AICore"),
+        .package(path: "../Networking"),
     ],
     targets: [
         .target(
             name: "AIProviders",
             dependencies: [
                 .product(name: "AICore", package: "AICore"),
+                .product(name: "Networking", package: "Networking"),
             ]
         ),
         .testTarget(
             name: "AIProvidersTests",
-            dependencies: ["AIProviders"]
+            dependencies: [
+                "AIProviders",
+                .product(name: "AICore", package: "AICore"),
+                .product(name: "Networking", package: "Networking"),
+                .product(name: "NetworkingMocks", package: "Networking"),
+            ]
         ),
     ],
     swiftLanguageModes: [.v6]
--- a/Modules/AIProviders/Sources/AIProviders/AIProviders.swift
+++ b/Modules/AIProviders/Sources/AIProviders/AIProviders.swift
@@ -1,2 +1,0 @@
-// The Swift Programming Language
-// https://docs.swift.org/swift-book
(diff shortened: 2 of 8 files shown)


n0an:
@claude please review this PR. It moves AnthropicService (the standalone Anthropic Messages API client) from the app target into the AIProviders module as the first concrete provider - should be a pure relocation (only public modifiers added). Focus on: (1) behavior preservation (HTTP headers, status-code mapping, JSON body, SSE parsing unchanged); (2) dependency direction (AIProviders -> AICore + Networking only, no app-type leak) and that the internal Logger+AIProviders extension can't collide with the app's; (3) the test move + AIProvidersTests deps + the new TestTags; (4) the 3 AIService call sites + import.

n0an:
@codex please review this PR. It relocates AnthropicService from the app into the AIProviders module (public struct), should be behavior-preserving. Focus on: any drift in the HTTP/SSE logic vs the original, correct module dependency direction with no app-only type leaking into the module, the internal (non-public) Logger extension not colliding with the app's, and the test/call-site wiring after the move.

PR #329, opened 2026-06-17
Move ReasoningConfig into AICore

n0an:
## Move ReasoningConfig into AICore

Relocates `ReasoningConfig` - a pure (Foundation-only) lookup of model → reasoning-effort parameters for the GPT-5 / Gemini / Cerebras / Groq families - from the app target into `AICore` as a `public` type.

### Why now

A prerequisite for the next provider move: `OpenAICompatibleService` calls `ReasoningConfig.getReasoningParameter(...)` / `getExtraBodyParameters(...)` when building request bodies, so the config has to live in the shared module before that service can move into `AIProviders`. Same pattern as the earlier `EnhancementError` / `AIEnhancementOutputFilter` foundational moves.

### Changes

- `ReasoningConfig` moves into `AICore`; the `struct` and its two `static func` entry points become `public` (the internal `static let` model-set tables stay internal - consumers only call the two functions).
- Adds a characterization test locking the mappings (`gpt-5.5` → `none`, `gpt-5-mini` → `minimal`, `gemini-2.5-flash` → `none`, Cerebras/Groq `gpt-oss` → `low`, unknown → `nil`, and the `zai-glm-4.7` `disable_reasoning` extra-body case).
- All three consumers (`OpenAICompatibleService`, `AIService+Chat`, `CloudReminderExtractionProvider`) already import `AICore`, so no import changes were needed.

### Behavior

Pure relocation - **no behavior change**, no `project.pbxproj` change.

### Verification

- `AICore` tests pass (`swift test`, now 3 suites).
- App + all three extensions + test bundle build green (`build-for-testing`, iOS 26.4 simulator).

### Next

Move the `OpenAICompatibleService` cluster (with `OllamaService` / `CustomOpenAIService`) into `AIProviders`.

### DIFF
--- a/Modules/AICore/Sources/AICore/ReasoningConfig.swift
+++ b/Modules/AICore/Sources/AICore/ReasoningConfig.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-struct ReasoningConfig {
+public struct ReasoningConfig {
     // 2.5-flash and 2.5-flash-lite support "none" to fully turn off thinking
     static let geminiNoneReasoningModels: Set<String> = [
         "gemini-2.5-flash",
@@ -60,7 +60,7 @@ struct ReasoningConfig {
         "qwen/qwen3-32b"
     ]
 
-    static func getReasoningParameter(for modelName: String) -> String? {
+    public static func getReasoningParameter(for modelName: String) -> String? {
         if geminiNoneReasoningModels.contains(modelName) { return "none" }
         else if geminiMinimalReasoningModels.contains(modelName) { return "minimal" }
         else if openAINoneReasoningModels.contains(modelName) { return "none" }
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR. It moves the pure ReasoningConfig (model -> reasoning_effort mappings) from the app into AICore as a public type - a prerequisite for moving OpenAICompatibleService. Focus on: behavior preservation (the mapping tables/functions byte-identical), the public-vs-internal surface (only the 2 get methods are public; the model-set static lets stay internal), and import coverage.

n0an:
@codex please review this PR. It relocates ReasoningConfig (model -> reasoning_effort lookup) from the app into the AICore module (public). Should be behavior-preserving. Focus on any drift in the model->parameter mappings, correct public/internal visibility, and whether the characterization test actually pins the mappings.

PR #333, opened 2026-06-17
Add AITextProvider protocol and per-provider wrappers

n0an:
## Add AITextProvider protocol + per-provider wrapper conformances

Introduces a uniform provider abstraction so the upcoming orchestrator/registry can treat every backend as `any AITextProvider` instead of a `switch` over provider types. This is the first **design** step (not a relocation) - the approach (wrapper structs, enhance-only scope) was chosen deliberately given the providers' divergent signatures.

### What

- **`AICore`: the `AITextProvider` protocol** - `enhance(systemMessage:userMessage:)` + `enhanceStreaming(...:onPartialResponse:)` (the partial callback is `@MainActor`). Foundation-only, so `AICore` stays a pure kernel.
- **`AIProviders`: six thin wrapper structs** - `AnthropicTextProvider`, `OpenAICompatibleTextProvider`, `AppleFMTextProvider`, `GeminiTextProvider`, `CopilotTextProvider`, `OpenAIOAuthTextProvider`. Each binds its backend's per-request config (API key / endpoint / headers / model / OAuth token / sampling profile) and delegates to the unchanged underlying service.

### Why wrappers (vs refactoring the services)

The 8 providers have genuinely divergent shapes - different config params, two naming conventions (`systemMessage/userMessage` vs `systemPrompt/text`), and static-`enum` vs instance forms. Thin wrappers bind the config and normalize naming at the boundary, leaving the already-extracted, already-tested services untouched. The composition root will build the right wrapper per request (next phase).

### Two details worth noting

- **OAuth naming map:** the Gemini/Copilot/OpenAIOAuth wrappers map `systemMessage → systemPrompt` and `userMessage → text`. Verified not swapped (a swap would silently send user text as the system prompt).
- **AppleFM isolation:** `AppleFoundationModelService` is `@MainActor`, so `AppleFMTextProvider` is `@available(iOS 26, macOS 26)` and constructs the service *inside* its async methods (which hop to the main actor) rather than in its nonisolated `init`.

### Behavior

**Purely additive - no behavior change.** Nothing is wired to these types yet (`git diff --stat HEAD` touches no existing file); they're introduced here and consumed by the registry in the next phase.

### Verification

- `AICore` 15 tests, `AIProviders` 70 tests (incl. wrapper delegation smoke tests), all green; both build under Swift 6 strict concurrency.
- App + all three extensions + test bundle build green (`build-for-testing`, iOS 26.4 simulator).

### Next

`AIProviderRegistry` in `AIKit` (its first real content) - maps a provider id to an injected `any AITextProvider` - then `AIService`'s enhancement path routes through the registry (the god-class shrink).

### DIFF
--- a/Modules/AICore/Sources/AICore/AITextProvider.swift
+++ b/Modules/AICore/Sources/AICore/AITextProvider.swift
@@ -1,0 +1,21 @@
+import Foundation
+
+/// A provider that turns a system message + user message into enhanced text,
+/// with an optional streaming variant that reports partial results.
+///
+/// Conformers bind their per-provider configuration (API key, endpoint URL,
+/// headers, model, OAuth token, sampling profile, ...) at construction, so this
+/// protocol stays uniform across every backend. Consumers (the orchestrator /
+/// registry) depend only on `AITextProvider` and never on a concrete provider.
+public protocol AITextProvider: Sendable {
+    /// Non-streaming enhancement. Returns the final enhanced text.
+    func enhance(systemMessage: String, userMessage: String) async throws -> String
+
+    /// Streaming enhancement. Calls `onPartialResponse` on the main actor as
+    /// partial text arrives, and returns the final enhanced text.
+    func enhanceStreaming(
+        systemMessage: String,
+        userMessage: String,
+        onPartialResponse: @escaping @MainActor (String) -> Void
+    ) async throws -> String
+}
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review. Adds the AITextProvider protocol (AICore) + 6 thin wrapper structs (AIProviders) that bind per-provider config and delegate to the unchanged services. Purely additive (nothing wired yet). Focus on: (1) the OAuth wrappers' naming map - confirm systemMessage->systemPrompt and userMessage->text are NOT swapped in Gemini/Copilot/OpenAIOAuth; (2) every wrapper delegates to the correct method with correct config binding; (3) AppleFMTextProvider's @available + building the @MainActor service inside async methods is correct (no deadlock); (4) AICore stays pure and the protocol/Sendable design is sound.

n0an:
@codex please review. Introduces AITextProvider (AICore) and 6 wrapper structs (AIProviders) delegating to the existing services, binding config and normalizing the systemMessage/userMessage vs systemPrompt/text naming for the OAuth providers. Additive only. Focus on any swapped/mis-bound argument in the 12 wrapper methods, the AppleFM @MainActor/availability handling, Sendable correctness, and that AICore stays dependency-free.

PR #334, opened 2026-06-17
Add AIProviderRegistry to AIKit (Phase 11a, additive)

n0an:
## What

Introduces the AI stack's composition point in `AIKit`:

- **`AIProviderRoute`** - the resolved-route enum (`appleFoundationModel`, `anthropic`, `cloud`, `openAIOAuth`, `geminiOAuth`, `copilot`, and the locally-configured `openAICompatibleEndpoint` for Ollama/Custom). The local-endpoint case carries an explicit `timeout` so self-hosted endpoints keep their longer request timeout (120s) instead of inheriting the cloud `baseTimeout` (300s).
- **`AIProviderRegistry`** - a `@MainActor struct` (the injected `OAuthManager` / `CopilotOAuthManager` are non-`Sendable`). `makeTextProvider(for:model:)` switches on the route, reads the API key via the injected `KeychainService` (mirroring `AIProvider.apiKey`), and retrieves OAuth tokens via `validAccessToken` / `validCopilotToken`. A missing key throws `EnhancementError.notConfigured`.

`AIKit` now depends on `AICore` + `AIProviders` + `Keychain` + `OAuth` + `Networking`, mirroring how `TranscriptionKit` depends on its impls + the infra they need.

## Why

This is the registry that the upcoming wiring phases will route enhancement through. It owns the per-backend config assembly that `AIService` currently does inline, so the next phases can replace that inline construction with a single `makeTextProvider` call.

## Scope

**Purely additive.** `AIService` is unchanged - it is not yet wired to the registry (that lands in the following phases). The app builds and the registry replicates `AIService`'s existing per-route config exactly.

## Tests

9 tests (`KeychainMocks` / `OAuthMocks` / `NetworkingMocks`) assert each route builds the correctly-configured provider:
- Anthropic and cloud routes seed the Keychain under the **real key name**, so a key-name swap fails the test; assertions check the outgoing endpoint, auth header, and bound model.
- OAuth routes assert the **correct OAuth provider per route** (a Gemini/OpenAI swap fails) and that token errors propagate.
- The local-endpoint route asserts the caller-supplied URL, headers, and **forwarded timeout** reach the request.
- A missing API key throws `EnhancementError.notConfigured`.

`swift test --package-path Modules/AIKit` and the full app build are green.

### DIFF
--- a/Modules/AIKit/Package.swift
+++ b/Modules/AIKit/Package.swift
@@ -12,20 +12,36 @@ let package = Package(
         .library(name: "AIKit", targets: ["AIKit"]),
     ],
     dependencies: [
-        // AIKit depends on the AICore API only - never on AIProviders (the impl).
-        // The composition root injects concrete providers into AIKit's registry.
+        // AIKit is the AI orchestration module: it builds and routes providers,
+        // so it depends on the provider impls + the infra they need (mirroring
+        // how TranscriptionKit depends on Cloud/Local transcription).
         .package(path: "../AICore"),
+        .package(path: "../AIProviders"),
+        .package(path: "../Keychain"),
+        .package(path: "../OAuth"),
+        .package(path: "../Networking"),
     ],
     targets: [
         .target(
             name: "AIKit",
             dependencies: [
                 .product(name: "AICore", package: "AICore"),
+                .product(name: "AIProviders", package: "AIProviders"),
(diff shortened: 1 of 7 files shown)


n0an:

n0an:
@codex review

PR #335, opened 2026-06-17
Wire AIService non-streaming enhance to AIProviderRegistry (Phase 11b)

n0an:
## What

Replaces the four inline provider-construction branches in `AIService`'s non-streaming `makeRequest` with `AIProviderRegistry.makeTextProvider(for:model:)` + `provider.enhance(...)`. `AIService` no longer assembles per-backend config itself for these routes:

- **default cloud switch** (Anthropic Messages API + OpenAI-compatible `Bearer`) → `.anthropic` / `.cloud(provider)`
- **OpenAI OAuth** happy-path → `.openAIOAuth`
- **Gemini OAuth** happy-path → `.geminiOAuth`
- **Copilot** happy-path → `.copilot`

A private `aiProviderRegistry` computed property is built from `AIService`'s injected dependencies (`keychain`, `oauthManager`, `copilotOAuthManager`, `networkService`, `logger`, `baseTimeout`).

## Behavior preserved

Each rewritten branch keeps its exact surrounding logic:
- the OAuth → CLI-server / API-key **fallback** (the `OAuthError` from the registry's `validAccessToken` still propagates out of `makeTextProvider` into the same `catch`)
- the Gemini **rate-limit `EnhancementError` rethrow** that must precede the OAuth fallback
- output filtering (`AIEnhancementOutputFilter`) and the `lastSystemMessageSent` / `lastUserMessageSent` side-effect ordering

## Deliberately unchanged

Not `AITextProvider` routes (or behavior would shift), so left exactly as-is:
- **Apple Foundation Model** — uses a *cached* `appleFoundationModelService` instance; the registry's wrapper builds a fresh one per call
- **Ollama / Custom OpenAI** — separate `makeOllamaRequest` / `makeCustomOpenAIRequest` methods
- the three **VivAgents CLI-server** branches

## Tests

- New `AIServiceEnhanceRoutingTests` characterizes the **Gemini OAuth** route (the one fully-mockable non-streaming path) end-to-end: a signed-in Gemini mode fetches the *Gemini* OAuth token and sends `Bearer <token>` to the cloudcode endpoint. Written **test-first** — green before and after the wiring.
- 84 AIService-dependent tests + the full `VivaDictaTests` suite + full app build all green.

(The Keychain-keyed cloud routes aren't characterized here because `AIProvider.apiKey` reads a file-private `DefaultKeychainService`, not the injected one — those are covered by `AIProviderRegistry`'s own tests.)

## Note

The registry's `requireAPIKey` adds a `!isEmpty` precondition the old inline guard lacked, so a stored **empty-string** key now throws `.notConfigured` before any network call instead of round-tripping to a 401. This is not reachable through the app's own flows (keys are only saved after verification; clearing deletes the key rather than storing `""`), and failing fast is strictly better on the only path that could reach it (external keychain corruption).

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -16,6 +16,7 @@ import OAuth
 import CloudTranscription
 import AICore
 import AIProviders
+import AIKit
 
 /// Service responsible for AI-powered text enhancement of transcriptions.
 ///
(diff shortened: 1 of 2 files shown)


n0an:

n0an:
@codex review

n0an:
@claude the previous review run ended before posting findings - please review this PR again.

PR #336, opened 2026-06-17
Wire AIService streaming enhance to AIProviderRegistry (Phase 11c)

n0an:
## What

Completes the AI request layer's move onto `AIProviderRegistry`. Replaces the five streaming `switch resolvedStreamingRoute` cases in `makeStreamingRequest` that built a provider/client inline with `AIProviderRegistry.makeTextProvider(for:model:)` + `provider.enhanceStreaming(...)`:

- `.anthropic` → `.anthropic`
- `.openAICompatibleCloud` → `.cloud(provider)`
- `.copilot` → `.copilot`
- `.geminiOAuth` → `.geminiOAuth`
- `.openAIOAuth` → `.openAIOAuth`

This is the streaming counterpart to #335 (non-streaming). After it, both request paths source their `AITextProvider`s from the registry.

## Behavior preserved

- **Filter layering** is kept exactly: `finalizeStreamingResult` (trim + `AIEnhancementOutputFilter` + re-emit) stays on **Copilot/Gemini/OpenAI-OAuth** (whose clients return raw text) and is **not** added to **Anthropic/cloud** (those services filter internally).
- The **OAuth → CLI / API-key fallback** (`catch OAuthError { … break }` → post-switch `makeRequest`) and the **Gemini rate-limit `EnhancementError` rethrow** that must precede the OAuth fallback are unchanged — the `OAuthError` from the registry's `validAccessToken` still propagates out of `makeTextProvider` into the same `catch`.
- Cloud streaming keeps `Authorization: Bearer <key>`, `baseTimeout`, and `errorPrefix: "<displayName> error"` (the registry's `.cloud` `errorPrefix` was aligned to this in the registry PR).

## Cleanup

- Removed the now-orphaned `makeAnthropicStreamingRequest` helper.
- `makeOpenAICompatibleStreamingRequest` stays — Ollama/Custom still use it.

## Deliberately unchanged

`.apple` (cached service instance), `.ollama` / `.customOpenAI` (separate request paths), and the CLI-server / `case nil` fallthrough.

## Tests

- New `geminiOAuthStreamingRouteSendsOAuthTokenToCloudCodeEndpoint` alongside the non-streaming one: a signed-in Gemini mode with an `onPartialResult` callback streams the Gemini OAuth token to the cloudcode endpoint. Written **test-first** — green before and after the wiring.
- The AIService suites + the full `VivaDictaTests` suite + the full app build are all green.

## Note

As in #335, the registry's `requireAPIKey` adds a non-empty precondition the old inline guard lacked, so a stored **empty-string** key now throws `.notConfigured` before any network call instead of issuing a guaranteed-401 request. Unreachable via app flows (keys are only saved after verification), and it makes streaming **consistent** with the already-shipped non-streaming registry path.

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -1065,23 +1065,6 @@ class AIService {
         OpenAICompatibleService.streamingDelta(from: line)
     }
 
-    private func makeAnthropicStreamingRequest(
-        systemMessage: String,
-        userMessage: String,
-        apiKey: String,
-        model: String,
-        onPartialResponse: @escaping @MainActor (String) -> Void
-    ) async throws -> String {
-        let service = AnthropicService(networkService: networkService, logger: logger, baseTimeout: baseTimeout)
-        return try await service.enhanceStreaming(
-            systemMessage: systemMessage,
-            userMessage: userMessage,
-            apiKey: apiKey,
-            model: model,
-            onPartialResponse: onPartialResponse
-        )
-    }
-
     private func makeStreamingRequest(
         text: String,
         mode: VivaMode,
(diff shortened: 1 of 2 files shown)


n0an:

n0an:
@codex review

PR #337, opened 2026-06-17
Make AIProviderRegistry the single source of the API-key precondition

n0an:
## What

Removes the three now-redundant cloud-path API-key guards in `AIService` so `AIProviderRegistry.requireAPIKey` is the **single source** of the API-key precondition:

- the non-streaming `makeRequest` cloud switch
- the streaming `.anthropic` case
- the streaming `.openAICompatibleCloud` case

Each was a `guard self.getAPIKey(for:) != nil else { throw .notConfigured }` immediately followed by `makeTextProvider(...)`, which calls `requireAPIKey` itself — so the key was read **twice, from two backends** (the guard via the file-private `DefaultKeychainService` inside `AIProvider.apiKey`; the registry via `AIService`'s injected `KeychainService`). In production both are the same `DefaultKeychainService`, so this only drops a redundant double-read.

This was flagged by the @claude reviews on #335/#336 and by the adversarial verification on each phase.

## Why it matters: it closes the cloud-route test gap

Because the cloud path no longer reads the file-private keychain via `AIProvider.apiKey`, a `MockKeychainService` can finally drive it end-to-end through `AIService`. New characterization tests:

- `cloudRouteSendsBearerKeyToProviderEndpoint` — OpenAI route reads the injected key and sends `Bearer <key>` to the OpenAI endpoint
- `anthropicCloudRouteSendsApiKeyHeaderToMessagesEndpoint` — Anthropic route sends `x-api-key` to `/v1/messages`
- `cloudRouteThrowsNotConfiguredWhenKeyMissing` — a missing key throws `.notConfigured` before any network call

These are the routes that previously *couldn't* be characterized through `AIService` (the doc comment in the test file used to say so).

## Behavior preserved

Adversarial verification (multi-agent): **PASS.**
- Missing key → identical `EnhancementError.notConfigured` (now from the registry).
- Present key → identical provider build / request.
- Same production keychain backend; only the redundant read is gone.
- The non-streaming `lastSystemMessageSent`/`lastUserMessageSent` are now assigned just before a missing-key throw instead of just after — **not observable** (they're only read on the success path to build a `TranscriptionVariation`).
- Scope is exactly the three guards: the OAuth/CLI fallback `getAPIKey != nil` checks, `resolvedStreamingRoute`, Apple/Ollama/Custom branches, and all debug logging are untouched.

The file-private `AIProvider.apiKey` stays for the OAuth/CLI **fallback-decision** checks — a separate concern from building the provider.

The full `VivaDictaTests` suite and the app build are green.

### DIFF
--- a/VivaDicta/Services/AIEnhance/AIService.swift
+++ b/VivaDicta/Services/AIEnhance/AIService.swift
@@ -1106,13 +1106,10 @@ class AIService {
                 throw EnhancementError.notConfigured
             }
         case .anthropic:
-            guard self.getAPIKey(for: aiProvider) != nil else {
-                throw EnhancementError.notConfigured
-            }
-
             logger.logDebug("AI Processing - Using Anthropic streaming")
             logger.logDebug("AI Processing - Model: \(mode.aiModel)")
 
+            // The registry reads the API key and throws `.notConfigured` if it's missing.
             let provider = try await aiProviderRegistry.makeTextProvider(for: .anthropic, model: mode.aiModel)
             return try await provider.enhanceStreaming(systemMessage: sysMsg, userMessage: userMsg, onPartialResponse: onPartialResponse)
         case .copilot:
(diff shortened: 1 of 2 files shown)


n0an:

n0an:
@codex review

PR #339, opened 2026-06-17
Introduce CLIServerEnhancer seam (Phase 12a)

n0an:
## What

First step of Phase 12 (lifting orchestration toward `AIKit`). Abstracts the app-level, all-static `VivAgentsClient` CLI server behind a protocol `AIKit` can depend on, so the enhancement orchestration depends on the *capability* rather than the concrete global.

- **AIKit:** `CLIServerProvider` enum + `@MainActor protocol CLIServerEnhancer` — `isCliActive(for:)` (folds in the master "CLI server enabled" toggle), `serverURL`, `enhance(text:systemPrompt:model:provider:)`. `@MainActor` because `VivAgentsClient`'s statics are main-actor-isolated (matches `AIProviderRegistry`).
- **App:** `DefaultCLIServerEnhancer` forwards to `VivAgentsClient`.
- **AIService:** injects `cliServerEnhancer` (default = the adapter) into both inits, and routes **every CLI orchestration call-site** through it: `resolvedStreamingRoute`'s three gates, the streaming + non-streaming OAuth→CLI fallback gates, and the three non-streaming CLI execution branches (incl. their `enhance` calls).

## Behavior-preserving

Each condition is replicated exactly — `isEnabled && isXCliActive` collapses to `isCliActive(for:)` (the adapter ANDs in `isEnabled`); the `serverURL` non-empty checks, the `provider:` strings (`anthropic`/`codex`/`gemini`), the fallback `|| getAPIKey != nil`, the filtering, and the `lastSystemMessageSent`/`lastUserMessageSent` ordering are unchanged.

The **config/connection-check** uses of `VivAgentsClient` (`refreshConnectedProviders`, the provider-availability checks) intentionally stay on the static — that's a different concern from request orchestration.

## Why it matters

The CLI fallback path (OAuth→CLI→API-key) was **untestable** because it reached a static. It's now mockable. Adds `MockCLIServerEnhancer` + a characterization test proving the Anthropic CLI route enhances via the injected enhancer.

## Also

Adds the `AIKit`/`AICore`/`AIProviders` test targets to the CI test plan, so the extracted module suites (registry tests, etc.) run in CI.

## Tests

New CLI characterization test + all routing tests + the full `VivaDictaTests` suite + the app build are green.

### DIFF
--- a/Modules/AIKit/Sources/AIKit/CLIServerEnhancer.swift
+++ b/Modules/AIKit/Sources/AIKit/CLIServerEnhancer.swift
@@ -1,0 +1,41 @@
+// Copyright © 2026 Anton Novoselov. All rights reserved.
+
+import Foundation
+
+/// The CLI backends an enhancement request can be routed through when a local
+/// CLI server (a Mac running the Claude / Codex / Gemini CLIs) is configured.
+public enum CLIServerProvider: String, Sendable {
+    case anthropic
+    case codex
+    case gemini
+}
+
+/// A boundary over the optional "CLI server" enhancement backend.
+///
+/// The enhancement orchestration can route a request through a user-configured
+/// local CLI server instead of (or as a fallback before) a hosted API. That
+/// server is an app-level concern (UserDefaults-backed config + auth), so this
+/// protocol lets the orchestration depend on the capability rather than the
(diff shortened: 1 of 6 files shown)


n0an:

n0an:
@codex review

PR #343, opened 2026-06-18
Drop redundant nonisolated keywords on already-nonisolated declarations

n0an:
## Summary

Removes **41 redundant `nonisolated` keywords**. The local SPM modules use the language-default (nonisolated) actor isolation; the app target is `@MainActor`-by-default (`SWIFT_DEFAULT_ACTOR_ISOLATION = MainActor`). A `nonisolated` keyword only does work when the enclosing context would otherwise be isolated - on a declaration that is *already* nonisolated it is pure noise (almost always carried over verbatim when code was extracted into modules).

### Removed (all on already-nonisolated declarations)

| File | Count | Why redundant |
|---|---|---|
| `AppGroup/AppGroupCoordinator.swift` | 27 | plain class in a nonisolated-default module - its methods, enums, `deinit`, and stored property were already nonisolated |
| `LiveTranslation/LiveTranslationAudio.swift` | 8 | members nested inside `nonisolated` helper classes (`PlaybackBufferQueue`, `AudioTeardownContext`, `TapInstaller`) |
| `AIEnhance/ExaWebSearchTool.swift` | 4 | `static func`s nested inside `nonisolated` enums (`ExaSearchClient`, `ExaAPIClient`) |
| `AudioPrewarmManager.swift` | 2 | members nested inside the `nonisolated AudioCaptureContext` class |

### Deliberately kept (load-bearing)

- **Type-level `nonisolated`** in the MainActor-default app target (`nonisolated struct ExaResult`, `nonisolated final class BackgroundTaskQueue`, the audio helper classes) - these opt the whole type out of the MainActor default.
- **File-scope `nonisolated` functions** (`installInputTapNonisolated`, `installLiveTranslationInputTap`) - global functions inherit the target's MainActor default too.
- **`nonisolated(unsafe)`** static state (AIProviders, DesignSystem, Exa's `networkService`) - a separate feature for mutable global state, unrelated to isolation defaults.
- **`@MainActor`/`actor` opt-outs** in `AudioRecording` (`@MainActor` class satisfying the nonisolated `AVAudioRecorderDelegate`) and `TranscriptionKit` (`actor`) - the type declares its own isolation, so the keyword is the required escape.

## Testing

- `xcodebuild ... -scheme VivaDicta -destination 'platform=iOS Simulator,name=iPhone 17 Pro Max,OS=26.4' build` -> **BUILD SUCCEEDED**

## Notes

No behavior change. Every removed keyword was on a declaration that was already nonisolated, so isolation semantics are identical - this is noise reduction only.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
+++ b/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
@@ -83,7 +83,7 @@ public final class AppGroupCoordinator: @unchecked Sendable {
     public static let kPendingLanguageOverride = "pendingLanguageOverride"
     public static let kPendingSourceTag = "pendingSourceTag"
 
-    nonisolated private enum UserDefaultsKeys {
+    private enum UserDefaultsKeys {
         static let isRecording = "isRecording"
         static let lastRecordingTimestamp = "lastRecordingTimestamp"
         static let transcribedText = "transcribedText"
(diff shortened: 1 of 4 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #349, opened 2026-06-19
Add model-graph cascade/nullify integration tests

n0an:
## Summary

Closes the highest-value integration-coverage gap: the **SwiftData delete-rule contract** across the model graph, pinned against a real in-memory store. Before this, only `Transcription` + `TranscriptionVariation` had a dedicated integration suite; the rest of the relationship/cascade web was untested at the integration level (and is invisible to unit tests).

`ModelGraphCascadeIntegrationTests` (3 tests) - each inserts the graph through one `ModelContext`, deletes through it, and re-reads through a **fresh** one (the Essential Developer "real store, separate instances" shape):

1. **`deletingTranscription_cascadesOwnedGraph_butNullifiesMultiNoteLink`** - deleting a `Transcription` cascades its *owned* graph (variations, tag assignments, chat conversations **+ their messages, two levels deep**) but only **nullifies** the many-to-many multi-note link, which survives.
2. **`deletingChatConversation_cascadesMessages_leavesTranscription`** - per-conversation cascade leaves the note untouched.
3. **`deletingMultiNoteConversation_cascadesMessages_butTranscriptionsSurvive`** - it owns its messages (cascade) but only references its transcriptions (many-to-many), which survive.

## Why
These cascade/nullify rules break **silently** when a model moves into a module or the CloudKit schema changes - exactly the kind of regression the modularization work risks. Unit tests can't see them. This makes the riskiest persistence surface safe to refactor.

## Testing
- 3 integration tests green (`.database`, `.integration` tags; reuses the existing in-memory `TestModelContainer` harness).

## Notes
- No production changes - tests only.
- Deliberately stops here rather than padding with more same-pattern cascade suites (Tags/Reminders/SmartSearch are the same delete-rule shape, now proven). Live e2e remains out of scope (16 providers = cost/flakiness/secrets).

### DIFF
--- a/VivaDictaTests/Integration/ModelGraphCascadeIntegrationTests.swift
+++ b/VivaDictaTests/Integration/ModelGraphCascadeIntegrationTests.swift
@@ -40,6 +40,9 @@ struct ModelGraphCascadeIntegrationTests {
 
         let assignment = TranscriptionTagAssignment(tagId: UUID(), transcription: transcription)
 
+        let reminder = ExtractedReminderDraft(title: "buy milk")
+        reminder.transcription = transcription
+
         let multiNote = MultiNoteConversation()
         multiNote.transcriptions = [transcription]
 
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #351, opened 2026-06-20
Tier B: deepen ViewModel tests + seam RecordViewModel for testability

n0an:
## Summary

Tier B of the test-coverage effort: raise app-target coverage by making ViewModels testable. Two kinds of change - **deepening tests on already-injectable chat ViewModels**, and **introducing dependency seams** so the heavily-coupled `RecordViewModel` can eventually be constructed in a test. Plus terminal coverage/LOC tooling.

All changes are behavior-preserving; every seam uses an empty-extension protocol conformance and a default that preserves existing call sites.

## ViewModel test deepening (no production change)

These VMs already took injectable protocols, so this is pure test work via the existing `MockAIChatService` + `MockAnalyticsService` + in-memory `ModelContainer`:

- **ChatViewModel** 4 -> 10 tests
- **SmartSearchChatViewModel** 3 -> 8 tests
- **MultiNoteChatViewModel** 3 -> 9 tests

Each covers the deterministic logic: `loadMessages` sorting, the four `sendMessage` guards (empty / whitespace / no-model / provider-not-ready), and `clearChat`. The big async paths (streaming replies, the full search-and-respond flow) are intentionally left uncovered - they route through statics + streaming.

## Dependency seams

- **`NoteSearchService`** - segregates the single `search` method `SmartSearchChatViewModel` uses from `RAGIndexingService.shared`; injected (default `.shared`), mockable.
- **`AIProcessingService`** - the mode-driven AI text-processing surface `RecordViewModel` uses (mode access, `isProperlyConfigured`, clipboard, presets, `enhance`, `generateVariation`). Empty `extension AIService: AIProcessingService`, same pattern as the existing `AIChatService`. Named "Processing" not "Enhance" - `enhance` is pre-AI-variations vocabulary. The wider reminder-extraction chain keeps concrete `AIService` (it needs provider config + auth), routed via `appState.aiService`.
- **`Transcriber` retype** - `RecordViewModel.transcriptionManager` now `any Transcriber` (the existing protocol already covers its 4 calls).
- **`AudioPrewarmer`** - the prewarmed-audio-session surface `RecordViewModel` uses; `AudioPrewarmManager` conforms, injected (default `.shared`).

Three of `RecordViewModel`'s construction blockers are now seamed (AI / transcription / audio). Two remain before a flow test is possible - the `AppState` god-object (its init boots the whole service graph incl. CloudKit) and `AppGroupCoordinator.shared` (wired in init). Those are a deliberate follow-up.

## Tooling

[...]

PR #353, opened 2026-06-20
Test RecordViewModel's transcribe->enhance->persist flow

n0an:
## Summary

Builds on the four dependency seams (merged via #351/#352) to add the **first real flow coverage** of `RecordViewModel` - the hardest-to-test ViewModel in the app, which until this arc could not even be constructed in a test. Drives `transcribeSpeechTask` end to end with all dependencies mocked and asserts the persisted result + the state pushed to the keyboard bridge.

## Tests added (`RecordViewModelTests`)

A `makeSUT` helper constructs the VM with a real `AppState()` + the four mocks (`AIProcessingService` / `Transcriber` / `AudioPrewarmer` / `AppGroupBridge`). A Parakeet mode skips the real-audio downsampling path so the mocked `transcribe` runs. Four `transcribeSpeechTask` tests:

- **transcribe + enhance**: the `Transcription` is persisted with both `text` and `enhancedText`, and `.transcribing` / `.enhancing` statuses are pushed to the bridge.
- **not configured**: enhancement is skipped; only the transcript persists.
- **enhancement fails**: the transcript is still persisted (text only) and an alert is raised, rather than losing the transcription.
- **no meaningful content**: a whitespace-only transcript is not persisted and state resets to idle.

These cover the main flow plus the catch block and the meaningful-content guard.

## Coverage

`RecordViewModel.swift`: **8.1% -> 40.7%** (the construction test had it at 8.1%; the flow tests cover `transcribeSpeechTask`'s ~150 lines and two error branches). The app-target headline moved **~7.0% -> 8.0%** - over a point from this one file, demonstrating the seam investment converting directly to coverage.

## Testing

`xcodebuild test -only-testing:VivaDictaTests/RecordViewModelTests` passes (10 tests).

### DIFF
--- a/VivaDictaTests/RecordViewModelTests.swift
+++ b/VivaDictaTests/RecordViewModelTests.swift
@@ -171,4 +171,46 @@ struct RecordViewModelTests {
         #expect(saved.first?.text == "just a note")
         #expect(saved.first?.enhancedText == nil)
     }
+
+    private enum FlowTestError: Error { case boom }
+
+    @MainActor
+    @Test func transcribeSpeechTask_whenEnhancementFails_persistsTranscriptAndAlerts() async throws {
+        let transcriber = MockTranscriber(currentMode: parakeetMode(), stubbedText: "kept transcript")
+        let ai = MockAIProcessingService()
+        ai.stubIsProperlyConfigured = true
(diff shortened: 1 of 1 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #359, opened 2026-06-26
Move the MLX on-device LLM runtime into the LocalLLM module

n0an:
## Summary

Consolidates both on-device LLM runtimes inside the `LocalLLM` module. The MLX glue and its `mlx-swift-lm` dependency move out of the app target into `LocalLLM`, alongside the LiteRT Gemma runtime it already owns - matching the module's documented charter (the heavy, iOS-only native framework is firewalled into the module instead of linked directly by the app, mirroring how `LocalTranscription` owns WhisperKit).

## Changes
- Moved `LocalMLXModel`, `LocalMLXModelManager`, `LocalMLXTextProvider` from `VivaDicta/Services/AIEnhance/MLX/` into `Modules/LocalLLM/Sources/LocalLLM/`, with the consumed API made `public`.
- `LocalLLM/Package.swift` now declares `mlx-swift-lm` (pinned to the same revision the app used) and links `MLXLLM` / `MLXLMCommon`.
- Removed `mlx-swift-lm` from the app target's direct dependencies; it now resolves transitively via `LocalLLM`.
- Moved `LocalMLXModelTests` into `LocalLLMTests` (it exercises internal members; already in the test plan); the app-side `MockLocalMLXModelEngine` now imports `LocalLLM`.
- Updated `documentation/Module-Architecture.md`: added the `LocalLLM -> MLX` edge + external node and rewrote the catalogue row (dropped the now-false 'MLX runtime stays app-side').

## Why
MLX was the lone on-device runtime living in the app target while LiteRT lived in `LocalLLM`. Moving it firewalls the heavy MLX/Metal framework behind the module boundary (the glue no longer recompiles on every app-code change) and makes the module consistent with its own purpose.

## Testing
- Builds clean (0 errors) for the iOS simulator.
- Confirmed `mlx-swift_Cmlx.bundle/default.metallib` is bundled into `VivaDicta.app` - mlx's Metal kernel library propagates through the module boundary, which was the original reason MLX was app-linked.
- Runtime MLX generation not yet exercised; recommend a quick check (download an MLX model and run a generation) to confirm Metal kernels load at runtime.

## Notes
No behavior change intended - pure relocation + access-level changes. The `mlx-swift-lm` revision pin is unchanged.

### DIFF
--- a/Modules/LocalLLM/Package.swift
+++ b/Modules/LocalLLM/Package.swift
@@ -13,10 +13,13 @@ let package = Package(
     ],
     dependencies: [
         .package(path: "../AICore"),
-        // On-device LLM runtime - owned here so the heavy, iOS-only native
-        // framework is firewalled into this module instead of linked directly
+        // On-device LLM runtimes - owned here so the heavy, iOS-only native
+        // frameworks are firewalled into this module instead of linked directly
         // by the app target (mirrors how LocalTranscription owns WhisperKit).
         .package(url: "https://github.com/n0an/swift-litert-lm.git", branch: "main"),
+        // MLX (GPU) runtime - pinned to the same revision the app target used
+        // before MLX was consolidated into this module.
+        .package(url: "https://github.com/ml-explore/mlx-swift-lm", revision: "edd42fcd947eea0b19665248acf2975a28ddf58b"),
     ],
     targets: [
         .target(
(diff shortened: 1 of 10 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #360, opened 2026-06-27
Add CoreML (ANE) on-device LLM + fix keyboard on-device AI crash/hang

n0an:
## Summary

On-device LLMs failed when AI processing was triggered from the **keyboard**: the keyboard delegates to the main app, which runs in the **background**, and iOS forbids Metal GPU command-buffer submission from a backgrounded app (`kIOGPUCommandBufferCallbackErrorBackgroundExecutionNotPermitted`). **MLX crashed** the process with an uncaught command-buffer error; **LiteRT/Gemma hung** in GPU kernel compile and left its load flag stuck, poisoning later requests. Transcription (ANE), Apple FM (out-of-process), and cloud were unaffected.

This PR fixes the crash/hang and adds a **CoreML runtime that runs on the Apple Neural Engine** - the one on-device LLM that works from the keyboard, since the ANE is permitted in the background.

## Changes

**Fix the GPU-in-background crash/hang**
- `AIService` refuses GPU on-device routes (LiteRT / MLX) when `applicationState == .background`, throwing an actionable error (`EnhancementError.onDeviceLLMRequiresForeground`) instead of hanging/crashing. The MLX crash is uncatchable after submission, so this prevents it.
- The keyboard pre-empts GPU on-device modes with an instant message rather than a doomed record→transcribe→fail round-trip.
- LiteRT and MLX managers self-heal a stuck load/generate flag (timestamped, 120s threshold; never clears an in-flight download).

**CoreML (ANE) on-device LLM - keyboard-capable**
- Add Qwen3.5 2B / 0.8B via `john-rocky/CoreML-LLM`, pinned to `.cpuAndNeuralEngine`. Runs on the ANE, so it works while backgrounded.
- `AIProvider.localRuntime(forModelID:)` classifies local ids by runtime (`-coreml` → ANE, `-mlx`/Gemma → GPU). CoreML is exempt from the foreground guard and keyboard pre-empt.
- **Lean system prompt for the CoreML path only** - CoreML uses recurrent (per-token) prefill, so a ~500-token system prompt added tens of seconds of time-to-first-token; the lean (~120-token) prompt cuts that. GPU runtimes batch-prefill and keep the full prompt.
- Download cards + model-picker wiring for the CoreML models (cards at the end of the Local tab, Qwen icon, "CoreML · Neural Engine" badge).

**Memory**
- Single-resident across the three runtimes: loading one unloads the others, so at most one ~1-2 GB model is in RAM.
- Free models under memory pressure (`DispatchSource`) and on background (unless a keyboard session is active), to avoid jetsam while holding a large model.

**UI**
- On-device download cards show an indeterminate "Downloading…" + spinner instead of a non-linear, stuck-looking percentage (multi-file model progress jumps then sits on the big weights file).

## Testing
- `xcodebuild` build succeeds (iPhone 17 Pro Max, iOS 26.4 simulator).
- On-device verified on a physical device: CoreML Qwen runs from the keyboard (background) and completes; MLX/LiteRT now fail fast with a clear message instead of crashing/hanging; downloads show the indeterminate state.

## Notes
- New SPM deps in `Modules/LocalLLM/Package.swift`: `CoreML-LLM` 1.9.0 + `swift-transformers`. The module is set to Swift 5 language mode (CoreML-LLM exposes non-Sendable `@Observable` generators with a token callback); its public API stays `Sendable`.
- Follow-up: Gemma 4 E2B/E4B CoreML (a different `Gemma4StatefulEngine`, separate pass).

### DIFF
--- a/Modules/AICore/Sources/AICore/AIProvider.swift
+++ b/Modules/AICore/Sources/AICore/AIProvider.swift
@@ -625,11 +625,49 @@ public enum AIProvider: String, CaseIterable, Identifiable, Codable, Sendable {
         case .customOpenAI:
             return [] // Model is configured by user
         case .local:
-            // Gemma (LiteRT) + MLX models; AIService picks the runtime by id.
+            // On-device models; AIService picks the runtime by id (see
+            // `localRuntime(forModelID:)`):
+            //   gemma-4-*   -> LiteRT (Metal GPU, foreground-only)
+            //   *-coreml    -> CoreML (Apple Neural Engine, works backgrounded)
+            //   *-mlx       -> MLX    (Metal GPU, foreground-only)
             return ["gemma-4-E2B", "gemma-4-E4B",
+                    "qwen3.5-2b-coreml", "qwen3.5-0.8b-coreml",
                     "qwen3.5-4b-mlx", "qwen3.5-2b-mlx", "qwen3.5-0.8b-mlx", "phi-4-mini-mlx",
                     "llama-3.2-1b-mlx", "llama-3.2-3b-mlx", "ministral-3b-mlx", "falcon3-3b-mlx", "granite-3.3-2b-mlx"]
         }
+    }
+}
+
+// MARK: - On-device runtime classification
+
+public extension AIProvider {
+    /// The on-device runtime backing a `.local` model id.
+    enum LocalRuntime: Sendable {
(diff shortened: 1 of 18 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #362, opened 2026-07-11
Refresh cloud transcription model lineup across five providers

n0an:
## Summary

Updates the cloud transcription catalog after a sweep of all 12 provider lineups (July 2026). One required migration and four new models:

### Soniox: stt-async-v4 -> stt-async-v5 (required)
- Soniox retired `stt-async-v4` on 2026-06-30; requests auto-route to v5 server-side, but the catalog, provider default, and AICore default now reference `stt-async-v5` directly.
- New `renamedCloudModels` map in `TranscriptionModelProvider` + `normalizedCloudModelName()` applied in `TranscriptionManager.getCurrentTranscriptionModel()`, so modes with the old v4 slug saved keep resolving instead of failing the catalog lookup.

### Gladia: Solaria-3 (new accuracy option)
- Async-only model for EN/FR/DE/ES/IT, notably more accurate on real-world business audio. Requires exactly one language, so its catalog entry excludes auto-detect.
- `GladiaTranscriptionService.Config` gains `modelName` (default `solaria-1`, unchanged behavior) and the create request now sends the `model` field. Solaria-1 keeps the recommended badge and stays the general default.

### Speechmatics: Melia-1 + API migration
- New multilingual model with automatic code-switching across 56+ languages at the lowest price tier.
- The job request migrates from the deprecated `operating_point` property to the new `model` property; the internal `speechmatics-batch-v2` name maps to `"enhanced"`. Melia-1 sends `language: "multi"` (required) and passes a user-picked language as `language_hints`.

### OpenAI: GPT-4o Transcribe Diarize
- Speaker-aware batch model that pairs with the Speaker Labels toggle. The service handles its distinct parameter set: `chunking_strategy=auto` (required for audio over 30s), no `language`/`temperature` fields, and `diarized_json` response format with speaker-turn formatting via `SpeakerDiarizationFormatter` when diarization is on. Catalog entry is auto-detect only since the model rejects a language parameter.

### Groq: Whisper Large V3
- Full-size Whisper as a higher-accuracy tier alongside Turbo (~10.3% vs ~12% WER). Catalog entry only; the service already passes model names through.

## Testing

- `swift test` in `Modules/CloudTranscription`: 200 tests, all passing.
- New tests: Gladia `model` field (default + solaria-3 with pinned language), Speechmatics enhanced/melia-1 request shapes (including no `operating_point` and hint omission on auto), OpenAI diarize request fields and diarized/plain response handling with consecutive-segment merging.
- Full app builds cleanly for iOS Simulator.

## Notes

- No breaking changes. Existing user selections are preserved; the only behavioral change for existing users is Soniox modes resolving to v5, which is the model Soniox already routes v4 requests to.
- Melia-1 is in production preview for batch on Speechmatics' side.

### DIFF
--- a/Modules/AICore/Sources/AICore/AIProvider.swift
+++ b/Modules/AICore/Sources/AICore/AIProvider.swift
@@ -404,7 +404,7 @@ public enum AIProvider: String, CaseIterable, Identifiable, Codable, Sendable {
         case .openRouter:
             return "openai/gpt-oss-120b"
         case .soniox:
-            return "stt-async-v4"
+            return "stt-async-v5"
         case .gladia:
             return "solaria-1"
         case .speechmatics:
--- a/Modules/CloudTranscription/Sources/CloudTranscription/GladiaTranscriptionService.swift
+++ b/Modules/CloudTranscription/Sources/CloudTranscription/GladiaTranscriptionService.swift
@@ -6,8 +6,9 @@ import os
 import TranscriptionCore
 
 /// Pre-configured Gladia STT client. Uses an upload + create + poll flow.
-/// `modelName` is unused by the v2 API (the endpoint always uses the default
-/// pre-recorded model), so the config omits it.
+/// The v2 API selects the speech model via the `model` field: `solaria-1`
+/// (default, 100+ languages, code-switching) or `solaria-3` (async-only,
+/// EN/FR/DE/ES/IT, exactly one language required - no code switching).
 public struct GladiaTranscriptionService: TranscriptionService, Sendable {
     private let logger = Logger(cloudTranscriptionCategory: "GladiaTranscription")
     private let apiBase = "https://api.gladia.io/v2"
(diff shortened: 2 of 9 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Review feedback addressed in 3d0d3c02f:

**Codex P2 (Melia-1 + translation)** - Confirmed against the Speechmatics language docs: translation is Enhanced/Standard only. The service now drops a configured translation target for `melia-1` jobs instead of sending `translation_config`, and the response path stays in sync so the transcription still succeeds (with a log line noting the ignored target). Covered by a new test asserting no `translation_config` in the job body and a successful plain result.

**Claude finding #2 (legacy v4 display name)** - `getTranscriptionModelDisplayName` now normalizes renamed slugs, so a saved `stt-async-v4` selection shows the Soniox v5 display name in Settings.

**Claude question #1 (diarize model with Speaker Labels off)** - Verified in OpenAI's speech-to-text docs: `gpt-4o-transcribe-diarize` supports `diarized_json`, `json`, and `text` response formats, so the plain `json` path when the toggle is off is valid as written. No change needed.

The two minor suggestions (Melia-1 language-picker messaging, theoretical double-space on empty diarized segments) are intentionally left as-is.

PR #364, opened 2026-07-19
Add Whisper language detection scores to Transcription Detail

n0an:
## What

Surfaces WhisperKit's full language-identification distribution (the ~99-language softmax Whisper computes before decoding) for locally-transcribed notes, and shows it in a bottom sheet on the Transcription Detail screen.

- New **globe toolbar button** on Transcription Detail (visible only when scores exist) opens a sheet listing all languages sorted by score: localized name, code, percentage, progress bar (`.presentationDetents([.medium, .large])`).
- Scores are captured only for **local WhisperKit transcriptions with language = auto** - the language-ID pass is a cheap extra forward call to `WhisperKit.detectLanguage(audioPath:)` after the transcription succeeds. A detection failure is logged and never fails the transcription.

## Implementation notes

- `TranscriptionCore.TranscriptionServiceResult` gains optional `languageProbabilities: [String: Double]` (source-compatible: defaulted init param, `.plain`/`.speakerAttributed` unchanged).
- `Transcriber` protocol + `TranscriptionManager.transcribe` now return `TranscriptionServiceResult` instead of a bare `String`, so backend metadata survives the post-processing pipeline. Call sites updated: `RecordViewModel` (incl. pending-save path), `WatchAudioProcessor` (watch notes also get scores), `AppState` wrapper (still returns `String`), retranscribe flow in `TranscriptionDetailView` (refreshes scores).
- SwiftData: new optional `languageProbabilities` attribute on `Transcription` - lightweight migration, CloudKit-compatible (optional Codable attribute). Older notes simply show no button.
- **Parakeet and all 14 cloud providers intentionally excluded**: FluidAudio's `ASRResult` exposes only text + confidence (language-code exposure is still an open FR, FluidInference/FluidAudio#303), and no cloud API returns a full language distribution (AssemblyAI/ElevenLabs return a single winning-language confidence at most).
- New `LanguageScore` value type keeps sorting/formatting testable; unit tests added for it and for the `TranscriptionServiceResult` field. Mock transcribers updated.

## Verification needed

Authored on a Linux box without Xcode - **no local build/test run was possible**. Please verify in CI or locally:
- `xcodebuild` for the app + watch targets
- VivaDictaTests (`LanguageScoreTests`, `RecordViewModelTests`, acceptance) and TranscriptionCore package tests
- A quick manual run: local WhisperKit model, language = auto → record → globe button on the detail screen

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Modules/LocalTranscription/Sources/LocalTranscription/WhisperKitTranscriptionService.swift
+++ b/Modules/LocalTranscription/Sources/LocalTranscription/WhisperKitTranscriptionService.swift
@@ -82,14 +82,27 @@ public final class WhisperKitTranscriptionService: @unchecked Sendable {
                 .joined(separator: " ")
                 .trimmingCharacters(in: .whitespacesAndNewlines)
 
+            let languageProbabilities = await detectLanguageProbabilities(
+                whisperKit: whisperKit,
+                audioURL: audioURL,
+                options: options
+            )
+
             if options.isSpeakerDiarizationEnabled,
                let diarizedResult = await makeSpeakerAttributedResult(from: result, audioURL: audioURL) {
                 logger.logNotice("✅ WhisperKit diarization completed successfully")
-                return diarizedResult
+                return TranscriptionServiceResult(
+                    text: diarizedResult.text,
+                    isSpeakerAttributed: diarizedResult.isSpeakerAttributed,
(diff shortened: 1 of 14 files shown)


n0an:
@claude take a look

PR #366, opened 2026-07-19
Preserve German "um" and add filler-word removal toggle

n0an:
## Summary

Fixes a user-reported issue: transcribing German speech with local models (Whisper Large Turbo, Parakeet V3) always removed the word "um" - a core German preposition ("um zu", "um 10 Uhr", "es geht um..."). The removal was not done by the models but by our own `TranscriptionOutputFilter`, which listed "um" as a universal filler stripped for every language.

### Part 1: Language-aware "um" handling

- "um" moved out of `universalFillers` into the per-language filler lists for `en`, `ru`, `es`, `fr` - languages where it is only a hesitation sound.
- German transcripts keep "um" while still stripping genuine German fillers ("äh", "ähm", "öh"...).
- Portuguese is fixed as a side effect: "um" is the Portuguese indefinite article ("um homem") and was previously stripped there too.

### Part 2: "Remove Filler Words" toggle

- New toggle in Settings -> Advanced -> Text Processing (default on), giving users full control to keep hesitation sounds verbatim in any language.
- `TranscriptionOutputFilter.filter()` gains a `removeFillers: Bool = true` parameter. When off, the filler pass (including language detection) is skipped entirely.
- Hallucination stripping (bracketed segments, tag blocks) runs regardless of the toggle - that is garbage removal, not style preference.
- New `isFillerRemovalEnabled` UserDefaults key, wired through `TranscriptionManager.transcribe()` using the same pattern as the existing `isReplacementsEnabled` gate.

## Testing

- Full `TranscriptionOutputFilterTests` suite passes on iPhone 17 Pro Max simulator: 35 tests, 0 failures.
- 8 new tests:
  - German "um zu" / "um 10 Uhr" survives while "ähm" is stripped
  - Portuguese "um homem" survives
  - Parameterized check that "um" is still stripped for en/ru/es/fr
  - Toggle off keeps fillers verbatim; hallucination brackets stripped either way
- Updated the old universal-fillers test to use "uhm" (still universal) instead of "um".

## Notes

- Behavior change beyond German/Portuguese: "um" was previously stripped for *every* language; it is now stripped only for `en`, `ru`, `es`, `fr`. For a language with no filler list (e.g. Italian, Dutch), an English "um" that spills into the transcript now survives rather than being removed. This is the intended safer tradeoff - keep a possibly-real word rather than delete it - and the toggle covers anyone who wants the old aggressive behavior off entirely.
- Auto-detect edge case: when the mode language is "auto", a very short German clip can still be misdetected as English by `NLLanguageRecognizer`, in which case "um" would be stripped. Setting the mode language explicitly to German is the reliable path.
- Pre-existing, out of scope: `"mm"` remains in the universal list and would strip a standalone "mm" in measurements like "10 mm".
- The filter runs only in the main app (`TranscriptionManager`); app extensions are unaffected.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
+++ b/Modules/AppGroup/Sources/AppGroup/UserDefaultsStorage.swift
@@ -95,6 +95,7 @@ public enum UserDefaultsStorage {
         public static let isChatEnabled = "isChatEnabled"
         public static let isLiveTranslationEnabled = "isLiveTranslationEnabled"
         public static let isStripTrailingPeriodEnabled = "isStripTrailingPeriodEnabled"
+        public static let isFillerRemovalEnabled = "isFillerRemovalEnabled"
 
         /// UUID string of the mode used for AI actions in note detail when the
         /// currently selected mode has no AI processing configured. Empty = none.
--- a/Modules/TextProcessing/Sources/TextProcessing/TranscriptionOutputFilter.swift
+++ b/Modules/TextProcessing/Sources/TextProcessing/TranscriptionOutputFilter.swift
@@ -17,10 +17,14 @@ public struct TranscriptionOutputFilter {
         #"\{.*?\}"#      // {}
     ]
 
-    /// Hesitation sounds that have no real-word collisions across the supported
-    /// languages. Always stripped regardless of language.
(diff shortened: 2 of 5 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #372, opened 2026-08-08
Add Grok subscription sign-in via OAuth device code

n0an:
## Summary

Lets users run Grok from a **SuperGrok or X Premium subscription** instead of a metered xAI API key.

The subscription access token is a plain bearer for `api.x.ai`, so it simply takes the API key's place on the existing OpenAI-compatible transport. No new client, no new wire format, and the model list is unchanged.

Sign-in uses the **device-code grant** rather than the browser redirect: xAI validates `redirect_uri` against the values registered for its public client, which would otherwise force a loopback listener on one specific port. Device code needs no callback at all.

## Device-code support is generic

Added to the OAuth module rather than to Grok, so the next device-code provider needs only a config struct:

- `OAuthProvider` gains optional `deviceCodeURL` / `deviceAuthExtraParams`. Redirect-only providers are unaffected.
- `OAuthManager` gains `startDeviceCodeFlow(provider:)` and `pollForDeviceCodeToken(provider:deviceCode:)`.
- `DefaultOAuthManager` implements RFC 8628:
  - **https-only verification URIs** - the URI is handed to the system opener, so a malformed or spoofed response must never be able to launch an arbitrary handler.
  - Strict validation of every response field.
  - Absent-or-zero `interval` floored to 5s (RFC 8628 permits `0`; a literal reading would produce a zero-delay poll loop).
  - `slow_down` honors the server-supplied interval.
  - `access_denied`, `authorization_denied` and `expired_token` handled distinctly; an unrecognized `error` code fails fast instead of spinning until the code expires.
- Credential construction extracted so the redirect, device-code and refresh paths share account extraction and expiry math.
- Form bodies encoded against RFC 3986's *unreserved* set. `.urlQueryAllowed` let reserved characters (`&`, `=`, `+`, spaces) through intact.
- `DeviceCodeResponse` gains `verificationUriComplete` plus a `browserURI` accessor preferring the pre-filled URI, so the user usually doesn't have to type the code.

## The 403 case is handled deliberately

xAI runs its own allowlist on the OAuth API surface, so sign-in can succeed while inference still answers **403** - reported by standard SuperGrok and X Premium subscribers alike. That case surfaces a message naming the cause and pointing at the API key, and **deliberately does not fall back silently**: spending the user's API credits without telling them is worse than an explanatory error.

Supporting it required widening `unauthorizedMessage` to cover 403 as well as 401, and adding the parameter to the non-streaming `enhance` path, which previously had no way to surface either. Ollama and Custom OpenAI get better 403 messages as a side effect.

## Incidental fix

`DefaultOAuthManager.shared` had no `backgroundTaskService`. Device-code polling continues while the user is in Safari approving the code, so the app is backgrounded for that stretch and polling would have been suspended. The Copilot manager already held an assertion; this one now does too.

## Testing

- App build **SUCCEEDED**; test targets build **SUCCEEDED**.
- **17 new device-code tests** covering the parsing and polling edge cases above (hostile URI schemes, missing/empty fields, `interval: 0`, `slow_down`, both denial codes, expiry, unknown error codes, credential persistence, default token lifetime).
- **2 new tests** pinning the 401/403 `unauthorizedMessage` mapping and confirming the generic mapping still applies without an override.
- Suites green: OAuth 42/42, AIProviders 109/109, AIKit 18/18, app `AIServiceEnhanceRoutingTests` + `AIServiceConfigurationTests` 37/37.

## Notes

- **Not verified end-to-end against a live subscription.** The device flow, token parsing and error mapping are unit-tested against synthetic responses, but the first real sign-in is still the real test.
- No fallback to the Grok Build surface (`cli-chat-proxy`) is included. It would need a second transport (Responses API plus client identity headers) that cannot be verified without a live subscription, so the 403 message stands in for it. Worth revisiting if real users hit the allowlist.
- The settings screen states up front that xAI limits which plans can connect an app, so a valid subscription can still be refused.

### DIFF
--- a/Modules/AIKit/Sources/AIKit/AIProviderRegistry.swift
+++ b/Modules/AIKit/Sources/AIKit/AIProviderRegistry.swift
@@ -57,6 +57,7 @@ public struct AIProviderRegistry {
         case .anthropic: AIProvider.normalizedModel(model, for: .anthropic)
         case .cloud(let provider): AIProvider.normalizedModel(model, for: provider)
         case .openAIOAuth: AIProvider.normalizedModel(model, for: .openAI)
+        case .grokOAuth: AIProvider.normalizedModel(model, for: .grok)
         default: model
         }
         switch route {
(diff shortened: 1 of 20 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #373, opened 2026-08-08
Support Base64/Data URI audio for custom transcription providers

### LINKED ISSUE #368: Support Base64/Data URI audio for custom transcription providers
[Polza.ai](http://polza.ai/) custom transcription returns HTTP 400

Hi! I’m trying to use [Polza.ai](http://polza.ai/) as a Custom OpenAI-compatible transcription provider.

Endpoint:
https://polza.ai/api/v1/audio/transcriptions

Polza expects the file field as a Base64 Data URI (data:audio/m4a;base64,...), while VivaDicta sends it as a binary multipart file. This results in HTTP 400.

The same endpoint/model/API key works in iOS Shortcuts when the audio is converted to Base64 first.

Would it be possible to support Base64/Data URI audio for custom transcription providers?

n0an:
Closes #368

## Summary

Custom OpenAI-compatible transcription endpoints were always sent a binary `multipart/form-data` upload. Some gateways only accept a JSON body whose `file` field is a Base64 data URI and answer multipart with HTTP 400, so those endpoints could not be used at all.

Each custom endpoint can now pick its wire format. Both formats carry the same fields (`file`, `model`, `language`, `response_format`, `temperature`) and share the existing response decoder, since the JSON path returns the same `{text, language, duration}` shape.

| Format | Content-Type | `file` field |
|---|---|---|
| `.multipartFormData` (default) | `multipart/form-data; boundary=…` | raw audio bytes as a file part |
| `.jsonBase64` | `application/json` | `data:<mime-type>;base64,<payload>` |

## Changes

- **`CustomTranscriptionRequestFormat.swift`** (new) - public enum in `CloudTranscription` with the two cases.
- **`CustomTranscriptionService.swift`** - `Config.requestFormat`; request building branches on it. `createRequestBody` renamed to `createMultipartBody`, new `createJSONBody` alongside it. The auto-vs-explicit language check is hoisted into one `explicitLanguage` property both builders share, so the two bodies cannot drift.
- **`CustomTranscriptionModel.swift`** - persists `requestFormat`. `init(from:)` is hand-written so configurations saved before the field existed decode as `.multipartFormData` rather than failing outright.
- **`CustomTranscriptionModelManager.swift`** / **`TranscriptionManager.swift`** - threaded through save and provider-config construction.
- **`AddCustomTranscriptionModelView.swift`** - segmented "Request Format" picker with a per-case hint ("Pick this if the server rejects file uploads").
- **`documentation/Transcription-System-Architecture.md`** - new "Custom Endpoint Request Formats" section.

## Testing

- 7 new tests in `CustomTranscriptionServiceTests` covering the JSON path: content-type, data-URI round-trip (the Base64 payload decodes back to the original file bytes), `model`/`response_format`/`temperature` fields, language present when explicit and omitted when auto, success path, and missing-file. Existing multipart tests are unchanged and still assert the default behaviour.
- `swift test --filter CustomTranscriptionServiceTests` - 18/18 pass.
- `xcodebuild build` (iPhone 17 Pro Max, iOS 26.4) - success, 0 errors.
- `xcodebuild test -only-testing:VivaDictaTests/TranscriptionManagerTests` - 7/7 pass.

## Notes

- No behaviour change for anyone who does not switch the picker; multipart stays the default for new and existing configurations.
- The data URI's MIME type comes from the existing `URL.audioMIMEType` helper, so iOS's WAV recordings send `data:audio/wav;base64,…` and Watch M4A sends `data:audio/mp4;base64,…`. Both subtypes are on the accepted file-type list of the usual upstream providers.
- Base64 inflates the payload by roughly a third. The whole file was already read into memory for the multipart path, so peak memory grows but the shape of the code does not.

### DIFF
--- a/Modules/CloudTranscription/Sources/CloudTranscription/CustomTranscriptionRequestFormat.swift
+++ b/Modules/CloudTranscription/Sources/CloudTranscription/CustomTranscriptionRequestFormat.swift
@@ -1,0 +1,19 @@
+// Copyright © 2026 Anton Novoselov. All rights reserved.
+
+import Foundation
+
+/// How `CustomTranscriptionService` puts the audio on the wire.
+///
+/// OpenAI's own `/audio/transcriptions` takes a binary `multipart/form-data`
+/// upload and most self-hosted Whisper servers copy that, so it stays the
+/// default. Some gateways instead expect a JSON body whose `file` field is a
+/// base64 data URI and answer multipart with HTTP 400, which is why the
+/// format is user-selectable per custom endpoint.
+public enum CustomTranscriptionRequestFormat: String, Codable, Sendable, CaseIterable {
+    /// `multipart/form-data` with the raw audio bytes as the `file` part.
+    case multipartFormData
+
+    /// `application/json` with the audio inlined into `file` as
+    /// `data:<mime-type>;base64,<payload>`.
+    case jsonBase64
+}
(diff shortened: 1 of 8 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

Issue #370, opened 2026-07-29
Use Soniox Realtime WebSocket for normal dictation (not just Live Translation)


aliafshany:
## Problem

Normal Soniox dictation uses the **async REST path** (upload → create job → poll every 1s → download), so transcription only starts **after** you stop speaking. That is much slower than realtime streaming (e.g. VoiceInk-style behavior).

Current pipeline:
```text
stop recording → upload full file → create job → poll (1s) → fetch transcript → cleanup → insert
```

## Finding

`SonioxRealtimeSTTClient` already exists and streams to `wss://stt-rt.soniox.com/transcribe-websocket` (PCM 16 kHz, partial/final tokens, endpoint detection) — but it appears wired only for **Live Translation**, not keyboard dictation.

Related: #369 (async model upgrade helps accuracy, not this latency architecture).

## Proposal

Offer **Soniox Realtime** for standard dictation:

1. Open WebSocket when recording starts
2. Stream PCM chunks while speaking
3. Accumulate final tokens
4. On stop, finalize and wait only for remaining tokens
5. Run existing cleanup / optional AI, then insert text
6. Support plain transcription mode (no `translation` object)
7. Use current realtime model (not a hardcoded older one)

### Smaller interim fix
Reduce async poll interval from **1s** to ~**200–300ms** (still won’t match streaming).

## Acceptance criteria

- [ ] Normal Soniox dictation can use realtime WebSocket streaming
- [ ] Transcript is largely ready at stop (no full-file upload+poll for this mode)
- [ ] Existing cleanup / AI post-processing still works
- [ ] Live Translation continues to work

Thanks!

n0an:
Hello
Thank you for your proposal. I'll implement this in the next version of the app.

aliafshany:
Unfortunately, still in the latest version, 3.8.0, the transcription is much slower compared to the translation part.

In translation, it is "happening in real time," but even though I have set my API customization to Soniox rt, it's still transcribing my conversation With huge delay. It is like grabbing my conversation and then uploading it, and then getting the results. But why is the translation function so fast?


Is there any test flight version so that if you try to fix this I could test it before publishing it

n0an:
I tested it, for me it works much faster when i select realtime soniox.
do you use in app or via keyboard? I haven't implemented speed boost for via keyboard path, bc it will be much more involved. So this realtime speedup will be noticeable only when you use in-app.

n0an:
[...]

PR #374, opened 2026-08-08
Stream Soniox dictation over the realtime WebSocket

n0an:
Addresses #370

## Problem

Soniox dictation used the async REST path - upload the finished file, create a job, poll until it completes - so transcription could not begin until the user stopped speaking. `SonioxRealtimeSTTClient` already existed but was wired only to Live Translation.

## Approach

A new **"Soniox Realtime (stt-rt-v5)"** entry in the transcription model picker. Selection is per-mode and the async entry stays available, so this is opt-in and the existing Soniox behaviour is untouched for anyone who does not pick it.

`AVAudioRecorder` exposes no buffers to stream, so streaming models get `StreamingAudioCapture` - an `AVAudioEngine` capture that writes the same 16 kHz mono WAV **and** fans the identical PCM out to the socket. The default recorder is not modified, so every non-streaming dictation on every provider takes exactly the path it did before.

```
realtime OFF  ->  AVAudioRecorder -> file -> upload/poll        (unchanged)

realtime ON   ->  AVAudioEngine ---+--> WAV file (fallback + storage)
                                   \--> 16k Int16 PCM -> WebSocket
```

Because the WAV is written throughout, a socket failure at any point falls back to the ordinary upload with the recording intact - which is why `finish()` throws rather than returning a partial transcript.

## Changes

- **`StreamingAudioCapture`** - engine capture writing the WAV plus a live PCM stream, bounded at ~1s of backpressure. Also exposes a mic level so the waveform keeps animating on this path.
- **`SonioxRealtimeDictationSession`** - actor owning the socket, accumulating tokens, and flushing on end-of-audio with a 10s tail guard.
- **`RealtimeTranscriptAccumulator`** - the merge rule as a pure value type: finals accumulate, non-finals *replace* the unstable tail. Getting that backwards duplicates every word as it firms up, so it is separated out and tested directly.
- **`RealtimeDictationCoordinator`** - pairs capture and session so `RecordViewModel` deals with one object.
- **`TranscriptionManager.postProcessStreamedText`** - streamed text runs through the same filters, formatting, replacements, and completion analytics as uploaded text. One implementation, so the two paths cannot drift.
- **`SonioxRealtimeSTTClient`** - gains a transcription-only mode (it previously always requested translation), language hints are now optional so auto-detect sends none, and the model moves off the deprecated `stt-rt-v4`.
- **`SonioxTranscriptionService`** - async polling ramps 200ms/500ms/1s instead of a flat 1s. Independent of streaming; helps the upload path on its own.

## Testing

- 7 new tests on the accumulator: finals accumulate, interims replace, interim-firming-into-final does not duplicate, empty batch clears the tail, mixed batches split correctly, translation tokens ignored.
- `VivaDictaTests` - 360/360 pass.
- `CloudTranscription` - 211/211 pass.
- `xcodebuild build` (iPhone 17 Pro Max, iOS 26.4) - success, 0 errors.

## Notes and caveats

- **The WebSocket round-trip is unverified.** Unit tests cover the merge rule and everything builds, but the live socket, the finalize handshake, and fallback-on-drop need a device with a Soniox key. Treat "streaming produces a correct transcript" as unproven until that is run.
- **Live Translation moves to `stt-rt-v5` as a side effect.** It was silently riding `stt-rt-v4`, deprecated 2026-06-30 and currently surviving only as an alias. This is a fix, but it does change Live Translation's model and is worth a look before shipping.
- **Keyboard dictation still uses upload-and-poll.** The prewarm capture has its own `AVAudioEngine` tap that is not wired to the socket yet, so issue acceptance criterion 1 is only met for the in-app record button. Left as follow-up rather than half-wired - hence "Addresses" rather than "Closes".
- `TranscriptionModelProvider.sonioxRealtimeModel` holds the model id because that file is compiled into the app extensions, which do not link the app-only realtime client.

### DIFF
--- a/Modules/CloudTranscription/Sources/CloudTranscription/SonioxTranscriptionService.swift
+++ b/Modules/CloudTranscription/Sources/CloudTranscription/SonioxTranscriptionService.swift
@@ -10,7 +10,6 @@ public struct SonioxTranscriptionService: TranscriptionService, Sendable {
     private let logger = Logger(cloudTranscriptionCategory: "SonioxTranscription")
     private let apiBase = "https://api.soniox.com/v1"
     private let maxWaitSeconds: TimeInterval = 300
-    private let pollIntervalNanoseconds: UInt64 = 1_000_000_000
 
     public struct Config: Sendable {
         public let apiKey: String
@@ -190,7 +189,19 @@ public struct SonioxTranscriptionService: TranscriptionService, Sendable {
                 throw CloudTranscriptionError.apiRequestFailed(statusCode: 504, message: "Transcription timed out")
             }
 
-            try await Task.sleep(for: .nanoseconds(pollIntervalNanoseconds))
+            try await Task.sleep(for: Self.pollInterval(elapsed: Date().timeIntervalSince(start)))
+        }
+    }
+
+    /// A flat 1s poll cost every dictation up to a second of dead air after the
+    /// job was already done. Short recordings - the common case - finish within
+    /// the first couple of seconds, so poll tightly there and back off for the
+    /// long files where an extra half-second is noise anyway.
+    private static func pollInterval(elapsed: TimeInterval) -> Duration {
+        switch elapsed {
+        case ..<2: .milliseconds(200)
+        case ..<10: .milliseconds(500)
+        default: .seconds(1)
         }
     }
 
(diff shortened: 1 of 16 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
@claude please review this PR again

n0an:
@codex please review this PR again

n0an:
@claude please review this PR again

n0an:
@codex please review this PR again

PR #375, opened 2026-08-09
Use whichever microphone is connected, consistently across every recording path

### LINKED ISSUE #371: AirPods stuck in low-quality headset mode after dictation ends (+ bonus items)
First off - thanks for VivaDicta. I use it on my iPhone every day, mostly through the keyboard, and it's genuinely become part of how I write.

There's one thing that's been bugging me for months, and while trying to work out whether it was the app or my AirPods, I found three adjacent things that seem worth flagging. They all come back to the same place - the persistent prewarm session - so I'm filing them as one issue rather than four. Happy to split them if you'd rather track them separately.

I went through the source to pin down where each one lives, and re-checked against current `main` a couple of days ago; all four still look present. I've referenced symbols rather than line numbers since those drift.

**Environment:** iPhone [model], iOS [version], AirPods [model], VivaDicta [version], keyboard extension.

---

## 1. AirPods stay in HFP after dictation finishes

The quality drop *during* recording makes sense - active mic means Bluetooth switches from A2DP to HFP. What doesn't make sense is that it stays that way afterwards. Music or a podcast keeps sounding like a speakerphone call until the keyboard session times out, or until some other app (Apple Dictation usually does it) takes the audio session and releases it.

**Steps to reproduce**

1. Connect AirPods and start playing audio.
2. Start a VivaDicta keyboard recording.
3. Stop the recording.
4. Go back to listening.

**Actual:** playback stays in low-quality headset mode.
**Expected:** the mic route is released once VivaDicta is done with it, and the AirPods return to A2DP.

From the code, `AudioPrewarmManager.startPrewarmSession()` configures the shared session with `.allowBluetoothHFP` and starts a continuous `AVAudioEngine` tap, and `stopRealCapture()` deliberately leaves the engine running - the comment there notes the timeout timer is intentionally deferred until processing completes. The session only actually gets deactivated in `endSession()`.

I understand why the session is persistent; that's what makes the keyboard flow feel instant, and I wouldn't want to lose that. The specific problem is holding the AirPods HFP route while idle. Two approaches that don't break the design: prewarm against the built-in mic (see item 4), or release/reconfigure the route once processing finishes and re-acquire it on the next recording.

## 2. No visible way to end the session early

Related to the above: as far as I can tell the only way to release the route immediately is the Live Activity's terminate control. Settings has a button to *start* the session ("Enable Keyboard Recording Session") which disables itself once active, but there's no matching "End session now" beside it. That would be a small addition and would have saved me a lot of guessing - for a while I assumed the muffled playback was an AirPods firmware thing.

## 3. Text-only keyboard actions still activate the mic

`vivadicta://activate-for-keyboard` logs itself as being for text processing, but it calls `AudioPrewarmManager.shared.startPrewarmSession()` - the same call the recording deep link `vivadicta://record-for-keyboard` makes. So a keyboard action that never records anything still spins up `AVAudioSession` and `AVAudioEngine` and flips the Bluetooth route. If that path doesn't need audio, a separate activation path that skips the session/engine setup would sidestep the whole problem for text actions.

## 4. Feature request: option to use the iPhone's built-in mic

There's no `setPreferredInput` call anywhere in the repo, so recording uses whatever input iOS picks - with AirPods connected, that's the AirPods. For dictation I'd nearly always rather use the phone mic: transcription is better, and playback stays in A2DP.

A two-option setting would cover it:

- **iPhone microphone** - sensible default for keyboard dictation
- **Automatic / connected device** - for people who specifically want AirPods, a wired headset, or a USB mic

Implementation-wise this is `setPreferredInput(_:)` with the `.builtInMic` port, after `setActive(true)` but before the engine and its tap are created. Worth noting it'd need threading through more than one call site: `AudioPrewarmManager` and `LiveTranslationAudio.configureSession()` configure the shared session independently (the latter uses `.allowBluetoothA2DP, .duckOthers` and no `.defaultToSpeaker`), so fixing one would leave the other inconsistent.

iOS 26 also ships `AVInputPickerInteraction`, a system input picker, if you'd rather surface device selection without building the UI.

## 5. Feature request: opt into iOS 26 high-quality Bluetooth recording

The flip side of item 4 - when someone *does* deliberately want to dictate through AirPods, iOS 26 offers a better route than plain HFP via the `.bluetoothHighQualityRecording` category option. Apple's guidance from WWDC25 ("Enhance your app's audio recording capabilities") is to add it alongside `.allowBluetoothHFP`; the system then prefers the high-quality route when the connected device supports it and falls back to HFP otherwise. Right now `[.mixWithOthers, .allowBluetoothHFP, .defaultToSpeaker]` is the only option set anywhere in the app.

With the iOS 18 deployment target it'd need gating:

```swift
var options: AVAudioSession.CategoryOptions = [
    .mixWithOthers,
    .allowBluetoothHFP,
    .defaultToSpeaker
]
if #available(iOS 26.0, *) {
    options.insert(.bluetoothHighQualityRecording)
}
```

Probably better to let the system decide support via the route rather than checking AirPods models - Apple listed AirPods 4, AirPods 4 with ANC, and AirPods Pro 2 at announcement, but that list will keep growing.

---

Of these, item 1 is the only one I'd really call a bug; the rest are things I noticed while chasing it. Item 2 would be a nice stopgap even if item 1 is a bigger change. Happy to test a build against my setup, or to put up a PR for the built-in-mic preference if that would be useful.

n0an:
Addresses #371, and fixes a separate user report that recording ignores a connected Bluetooth headset.

## Problem

Two complaints that look opposite but share one cause — every recording path configured the audio session by hand, and they had drifted apart:

- **A connected headset is ignored.** A user wearing Beats found VivaDicta recording through the iPhone mic. He isolated it well: walking away from the phone faded his voice out, and the same headset records correctly in Voice Memos. Reproduced on device. Cause: the in-app record path never requested `.allowBluetoothHFP`, so iOS had no permission to use a Bluetooth mic and fell back to the built-in one.
- **AirPods stay in headset audio** (#371 item 1). The keyboard path requested `.allowBluetoothHFP` unconditionally and held it, so music kept sounding like a speakerphone call until the session expired.

Same knob, no setting, and inconsistent between paths — the in-app button and the keyboard behaved differently for the same user.

## Approach

A **Microphone** setting, defaulting to **Automatic**:

| Option | Behaviour |
|---|---|
| **Automatic** (default) | Requests the Bluetooth input route, sets no preferred-input pin — iOS routes to whatever is connected, the way Phone and Voice Memos do. On iOS 26 adds `.bluetoothHighQualityRecording`, so supported AirPods take the high-quality route rather than HFP |
| **iPhone Microphone** | Pins the built-in mic and drops `.allowBluetoothHFP` entirely, so headphones never leave A2DP |

Measured against current behaviour this is no worse for anyone and better for most: in-app recording now uses a connected headset (the bug), keyboard dictation is unchanged rather than losing its headset, and on iOS 26 the audio-quality complaint is largely addressed rather than traded away.

An earlier revision of this PR defaulted to pinning the built-in mic. That fixed #371 item 1 outright but left the headset report broken until the user found a setting, and silently regressed keyboard dictation. The commit history has that reasoning.

## Where the policy lives

`RecordingAudioSession` sits in `AppGroup` rather than the app target, because `DefaultAudioRecordingService` — the app's primary recorder — is its own SPM module and cannot import app-target code. Four call sites configure the shared session; putting the policy below the lowest of them is what makes the coverage structural instead of a list someone has to remember:

- `AudioPrewarmManager` (keyboard / prewarm)
- `DefaultAudioRecordingService` (in-app record button)
- `StreamingAudioCapture` (realtime dictation)
- `LiveTranslationAudio` — takes the Bluetooth *input* options from the helper while keeping the A2DP output and `duckOthers` its TTS playback needs. `setPreferredInput` can only choose among devices the category permits, so omitting HFP there would have pinned the built-in mic regardless of the setting

Changing the setting while a keyboard session is live restarts it, so the choice applies immediately rather than at session expiry.

## Issue #371 item by item

| # | Status |
|---|---|
| 1 AirPods stuck in HFP | **Partly.** Not fixed by default any more — that trade cost more than it bought. On iOS 26 the high-quality route largely resolves it; otherwise "iPhone Microphone" or "End Session Now" |
| 2 End session early | Done — button in Settings while a session is active |
| 3 Text actions activate the mic | **Won't fix as specified**, see below |
| 4 Prefer built-in mic | Done — the setting, plus `setPreferredInput` at all four sites |
| 5 iOS 26 high-quality Bluetooth | Done — `.bluetoothHighQualityRecording`, gated on `#available(iOS 26.0, *)` |

**On item 3:** the issue asks that `vivadicta://activate-for-keyboard` stop calling `startPrewarmSession()`, since it only ever processes text. That reasoning doesn't survive checking why the call is there. The app declares the `audio` background mode, and the live `AVAudioSession` is what keeps the main app resident after the keyboard returns the user to the host app — which is how the text-processing request gets serviced at all. Removing it suspends the app and text actions silently stop working. It stays, with a comment at the call site recording why.

## Testing

- `xcodebuild build` (iPhone 17 Pro Max, iOS 26.4) — success, 0 errors.
- `VivaDictaTests` 366/366, `AudioRecording` 11/11.
- **Route behaviour is not covered by tests and cannot be** — A2DP retention, `setPreferredInput`, and the iOS 26 high-quality path need hardware. The headset report has been reproduced on device, so the specific case to verify is: Automatic + a Bluetooth headset + the in-app record button, walking away from the phone.

## Notes

- Behaviour change for in-app recording: it now uses a connected headset where it previously always used the phone mic.
- `.bluetoothHighQualityRecording` is gated against the iOS 18 deployment target.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
+++ b/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
@@ -292,16 +292,36 @@ public final class AppGroupCoordinator: @unchecked Sendable {
 
     // MARK: - Keyboard Dictation Session Management
 
+    /// Sentinel `timeoutSeconds` for a keyboard session that never expires on its
+    /// own (Settings -> Keyboard -> Session Timeout -> "Never"). Such a session is
+    /// stored with an expiry in the distant future, so every expiry comparison
+    /// keeps working unchanged and only an explicit
+    /// ``deactivateKeyboardSession()`` ends it.
+    public static let sessionTimeoutNever = 0
+
+    /// Wall-clock expiry stamp for a session started/refreshed right now with
+    /// `timeoutSeconds`, honouring the ``sessionTimeoutNever`` sentinel.
+    private func expiryTime(forTimeout timeoutSeconds: Int) -> TimeInterval {
+        guard timeoutSeconds != Self.sessionTimeoutNever else {
+            return Date.distantFuture.timeIntervalSince1970
+        }
+        return Date().timeIntervalSince1970 + Double(timeoutSeconds)
+    }
+
     /// Activates a keyboard recording session with the specified timeout.
     ///
-    /// - Parameter timeoutSeconds: How long the session remains active without activity.
+    /// - Parameter timeoutSeconds: How long the session remains active without
+    ///   activity, or ``sessionTimeoutNever`` for a session with no expiry.
     public func activateKeyboardSession(timeoutSeconds: Int) {
-        let expiryTime = Date().timeIntervalSince1970 + Double(timeoutSeconds)
         sharedDefaults?.set(true, forKey: UserDefaultsKeys.keyboardSessionActive)
-        sharedDefaults?.set(expiryTime, forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
+        sharedDefaults?.set(expiryTime(forTimeout: timeoutSeconds), forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
         sharedDefaults?.set(Date().timeIntervalSince1970, forKey: UserDefaultsKeys.lastRecordingTimestamp)
         postDarwinNotification(NotificationNames.keyboardSessionActivated)
-        logger.logError("🔑 Keyboard session activated for \(timeoutSeconds) seconds")
+        if timeoutSeconds == Self.sessionTimeoutNever {
+            logger.logError("🔑 Keyboard session activated with no timeout")
+        } else {
+            logger.logError("🔑 Keyboard session activated for \(timeoutSeconds) seconds")
+        }
     }
 
     public var isKeyboardSessionActive: Bool {
@@ -337,10 +357,13 @@ public final class AppGroupCoordinator: @unchecked Sendable {
         guard let defaults = sharedDefaults else { return }
         let isActive = defaults.bool(forKey: UserDefaultsKeys.keyboardSessionActive)
         guard isActive else { return }
-        let newExpiryTime = Date().timeIntervalSince1970 + Double(timeoutSeconds)
-        defaults.set(newExpiryTime, forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
+        defaults.set(expiryTime(forTimeout: timeoutSeconds), forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
         defaults.set(Date().timeIntervalSince1970, forKey: UserDefaultsKeys.lastRecordingTimestamp)
-        logger.logError("🔁 Keyboard session expiry refreshed for \(timeoutSeconds) seconds")
+        if timeoutSeconds == Self.sessionTimeoutNever {
+            logger.logError("🔁 Keyboard session expiry refreshed with no timeout")
+        } else {
+            logger.logError("🔁 Keyboard session expiry refreshed for \(timeoutSeconds) seconds")
+        }
     }
 
     // MARK: - Keyboard Clipboard Context
--- a/Modules/AppGroup/Sources/AppGroup/RecordingAudioSession.swift
+++ b/Modules/AppGroup/Sources/AppGroup/RecordingAudioSession.swift
@@ -59,12 +59,19 @@ public enum RecordingAudioSession {
     /// Note this is load-bearing in both directions: `setPreferredInput` can
     /// only choose among devices the *category* permits, so omitting HFP here
     /// silently pins the built-in mic no matter what the preference says.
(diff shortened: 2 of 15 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
@claude please review this PR again

n0an:
@codex please review this PR again

n0an:
@claude please review this PR again

n0an:
@codex please review this PR again

n0an:
@claude please review this PR again

n0an:
@codex please review this PR again

PR #376, opened 2026-08-09
Add a Never option to the keyboard session timeout

n0an:
## Problem

The **Settings -> Keyboard -> Session Timeout** picker topped out at **1 hour**, so a hot mic session always expired on a timer. Users who want the keyboard to stay ready indefinitely had no way to express that.

## Change

Adds a **Never** option to the picker, backed by the sentinel `AppGroupCoordinator.sessionTimeoutNever` (`0`).

`AudioPrewarmManager` branches on a single `isTimeoutDisabled` check:

| Site | Behavior with Never |
|---|---|
| `scheduleSessionTimeout()` | Invalidates any existing timer and schedules nothing |
| `isWithinSessionTimeout()` | `true` for any started session, so `startRealCapture()` cannot throw `.sessionExpired` |
| `timeoutRemaining` | `.infinity` |

`AppGroupCoordinator.activateKeyboardSession(timeoutSeconds:)` and `refreshKeyboardSessionExpiry(timeoutSeconds:)` store `Date.distantFuture` as `keyboardSessionExpiryTime` for the sentinel. That keeps the keyboard extension's `isKeyboardSessionActive` expiry comparison working unchanged - no keyboard-side code touched, and the existing "still recording keeps it alive" path is untouched too.

No call site of `audioSessionTimeout` needed changes; they all funnel through the two coordinator methods above.

## Ending a Never session

There is no wall-clock expiry, so the session ends only on an explicit `endSession()`:

- the Live Activity terminate action (`ToggleSessionIntent`), and
- the launch-time session reset when the app is relaunched.

The picker shows a note about this, plus the battery cost, while Never is selected.

## Tests

Three new cases in `AppGroupCoordinatorSessionTests` cover activation, refresh, and explicit deactivation under the sentinel.

Not built or run locally per request - CI/reviewers to verify.

## Docs

`Hot-Mic-Audio-Prewarm-Architecture.md` and `AppGroupCoordinator-Architecture.md` updated with the sentinel semantics.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
+++ b/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
@@ -292,16 +292,36 @@ public final class AppGroupCoordinator: @unchecked Sendable {
 
     // MARK: - Keyboard Dictation Session Management
 
+    /// Sentinel `timeoutSeconds` for a keyboard session that never expires on its
+    /// own (Settings -> Keyboard -> Session Timeout -> "Never"). Such a session is
+    /// stored with an expiry in the distant future, so every expiry comparison
+    /// keeps working unchanged and only an explicit
+    /// ``deactivateKeyboardSession()`` ends it.
+    public static let sessionTimeoutNever = 0
+
+    /// Wall-clock expiry stamp for a session started/refreshed right now with
+    /// `timeoutSeconds`, honouring the ``sessionTimeoutNever`` sentinel.
+    private func expiryTime(forTimeout timeoutSeconds: Int) -> TimeInterval {
+        guard timeoutSeconds != Self.sessionTimeoutNever else {
+            return Date.distantFuture.timeIntervalSince1970
+        }
+        return Date().timeIntervalSince1970 + Double(timeoutSeconds)
+    }
+
     /// Activates a keyboard recording session with the specified timeout.
     ///
-    /// - Parameter timeoutSeconds: How long the session remains active without activity.
+    /// - Parameter timeoutSeconds: How long the session remains active without
+    ///   activity, or ``sessionTimeoutNever`` for a session with no expiry.
     public func activateKeyboardSession(timeoutSeconds: Int) {
-        let expiryTime = Date().timeIntervalSince1970 + Double(timeoutSeconds)
         sharedDefaults?.set(true, forKey: UserDefaultsKeys.keyboardSessionActive)
-        sharedDefaults?.set(expiryTime, forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
+        sharedDefaults?.set(expiryTime(forTimeout: timeoutSeconds), forKey: UserDefaultsKeys.keyboardSessionExpiryTime)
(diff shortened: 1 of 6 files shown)


n0an:
@claude review pls

n0an:
Both reviews landed on the same P1/P2, and it's a real bug - thanks. Fixed in c89047e7b.

**Scope was actually one entry point wider than reported.** Two flows start a prewarm session without a Live Activity, not one:

- `activateKeyboardRecordingSession()` in Settings (the reported case)
- the `vivadicta://activate-for-keyboard` text-processing deeplink (`VivaDictaApp.swift:471`), which starts prewarm and activates the coordinator session but never calls `startLiveActivity()`

Only `vivadicta://record-for-keyboard` starts one (`VivaDictaApp.swift:424`). With a finite timeout both unhandled flows self-terminated on the expiry timer, which is why this never surfaced before.

That ruled out Option A: starting a Live Activity in `activateKeyboardRecordingSession()` would have patched the Settings flow and left the text-processing deeplink stranded. Went with **Option B** instead - the Settings row is now a start/stop control (`KeyboardSessionButton`), calling `endSession()` while `isSessionActiveObservable` rather than sitting `.disabled`. Because the state lives on the shared `AudioPrewarmManager`, it stops a session started by *any* entry point, and it's an explicit in-app stop for every timeout value, not just Never.

`endSession()` deactivates the shared keyboard session, so the keyboard is notified and any Live Activity is torn down via the existing `onKeyboardSessionExpired` handler - same teardown the timeout path uses.

Also retargeted the picker warning, which named a Live Activity this flow never creates.

On the minor notes: `timeoutRemaining` having no callers is correct, I'm leaving the `.infinity` branch in as the honest answer for a disabled timeout rather than deleting a public-ish property. And agreed on the `0`-sentinel maintenance note - `isTimeoutDisabled` is the single guard any future arithmetic should route through.

Not built locally (DerivedData state in this worktree makes a local build unreliable) - CI is the check.

n0an:
@claude review pls

n0an:
@codex review

n0an:
Codex's P2 is valid and it's worse than reported - fixed in 77c820216.

`scheduleSessionTimeout()` only runs at session start, on extend, and after processing, so a live session keeps whatever timer it started with. **Both** directions of a mid-session picker change are wrong:

- **finite → Never** (the reported case): the old timer still fires at its original deadline, contradicting the warning rendered directly above the picker.
- **Never → finite** (not reported, and worse): there is no timer to reschedule, so none gets scheduled - the microphone stays armed indefinitely, and the warning disappears from the screen because the sentinel no longer matches. Silently stranded, which is exactly the failure mode the previous commit set out to close.

`applyTimeoutChange()` reschedules the local timer and the shared keyboard expiry from the picker's `onChange`. It deliberately defers while a real capture is in flight: `startRealCapture()` invalidates the timer on purpose, and the `rescheduleSessionTimeout()` that runs after processing already reads the current `audioSessionTimeout`, so the change lands there instead. Self-healing rather than a special case.

@claude - on your minor note about **End** during an active capture: agreed, leaving it. Stopping a recording is what a stop control should do, and the alternative (disabling the row mid-capture) reintroduces a smaller version of the unreachable-stop problem. Worth knowing it's a behavior change beyond the Never case, which is why it's called out in the architecture doc rather than left implicit.

PR #380, opened 2026-08-29
Return to 15 more host apps, and stop losing mappings to the query cap

n0an:
## Summary

The keyboard hands its host app's bundle ID to the main app, which looks up a way back and sends the user there. Two problems with that lookup.

**The lookup was silently capped.** The scheme table had reached 50 distinct schemes and `LSApplicationQueriesSchemes` 52 entries. Apple caps that array at 50, and beyond the cap `canOpenURL` returns `false` regardless of whether the app is installed — so the gate in front of the open was failing for entries it had no opinion about, and the user fell through to the manual return prompt. The two most recently appended schemes (`gemini-app`, `yandextranslate`, both added in the previous mapping pass) were the likely casualties.

**The gate is removed rather than the array trimmed.** `LSApplicationQueriesSchemes` exists solely to permit `canOpenURL`; `open` requires no declaration and already reports failure through its result. The code was doing both anyway — checking, then opening and logging the outcome. It now just opens and falls back to the prompt when that fails, which removes the ceiling permanently. These were the only two `canOpenURL` calls in the app, so the array is dead and deleted with them.

## Behavior change

Recording now starts before the open is known to have succeeded. `open` only answers after it has tried, and the user arrives in the host app expecting to already be recording, so waiting would cost the head of the recording. The adjacent no-scheme branch already made that same trade.

## The 15 new mappings

Two were verified from primary sources rather than taken from a list:

- **Swiftgram** (`app.swiftgram.ios`, 23 of 65 events) registers `telegram`, `tg`, `tonsite` and `sg`. Only `sg` is its own — the rest are shared with official Telegram, and iOS picks unpredictably when two apps claim a scheme. Confirmed against the project's Bazel `BUILD` file and diffed against upstream Telegram-iOS.
- **Apple Reminders** registers exactly one scheme, `x-apple-reminderkit`, read out of the shipping binary in a simulator runtime. The `x-apple-reminder://` that circulates on blogs is not registered and would fail silently.

High confidence: `sg://`, `minis://` (Open Minis, from its own `Info.plist`), `weixin://` (WeChat, Tencent docs), `spotify://` for the team-ID-suffixed Spotify build, `x-apple-reminderkit://`, `letterboxd://` (official repo), `eudic://` (official docs).

Medium confidence — corroborated across independent sources but not read from a shipping app, and a miss degrades to the prompt: `notion://`, `imeituan://`, `nplayer-http://`.

**Universal-link fallbacks** for five apps that register no scheme but claim a link in their AASA file: YouTube Studio, Amazon DE, ivi, MAX, Ozon. This works here in a way it would not generally — the lookup only ever runs for the app the keyboard was just typing into, so it is installed by definition. Residual risk: a user who has told iOS to open that domain in Safari lands on the web page instead of the prompt.

## Other changes

- `getURLSchemeForBundleId` → `returnURL(forHostId:)`, now returning a `URL`. The old name stopped describing a table that also holds universal links.
- Five hosts added to `knownNoSchemeHosts` so they stop being reported, including `com.anton`, which matches no App Store app in any storefront and is a dev build.

## Testing

- Builds clean for the app and all extensions (0 errors, 0 new warnings).
- Mapping table validated programmatically: 66 entries, no duplicate bundle IDs, no malformed URLs.
- `Info.plist` passes `plutil -lint`; the diff is a pure deletion with no reformatting.
- The Swiftgram and Reminders schemes were each confirmed independently against their source of truth.
- Not exercised end-to-end on device — the return path needs a real host app per mapping.

## Note on the data

`.unrecognizedHostApp` only fires when a host ID resolves, and that returned `nil` on iOS 26.4+ for the entire reporting window. This sample is therefore iOS 25 and older users only: it under-reports, and the app mix will shift now that resolution works again. Worth re-running once 3.9.0 has been out a few weeks.

### DIFF
--- a/.agents/skills/analyze-unrecognized-apps/SKILL.md
+++ b/.agents/skills/analyze-unrecognized-apps/SKILL.md
@@ -6,7 +6,7 @@ disable-model-invocation: true
 
 # Analyze Unrecognized Host Apps
 
-**Last run:** 2026-03-29 — added 7 URL scheme mappings (ShellFish, Termius, Teams, Reddit, iA Writer, Gemini, Yandex Translate) and 12 knownNoSchemeHosts entries
+**Last run:** 2026-08-29 - added 15 return-URL mappings (Swiftgram, Open Minis, WeChat, Spotify team-ID variant, Reminders, Letterboxd, Eudic, Notion, Meituan, nPlayer, plus 5 universal-link fallbacks) and 5 knownNoSchemeHosts entries. Also removed the `canOpenURL` gate and the dead `LSApplicationQueriesSchemes` array, which had exceeded Apple's 50-entry cap and was silently killing the most recent mappings. Note: this sample predates the iOS 26.4 host-resolution fix, so it under-reports - re-run once 3.9.0 has been out a few weeks.
 
 You are given the following context:
 $ARGUMENTS
--- a/.agents/skills/analyze-unrecognized-apps/SKILL.md
+++ b/.agents/skills/analyze-unrecognized-apps/SKILL.md
@@ -17,40 +17,53 @@ Analyze unrecognized host app bundle IDs from Firebase/Google Analytics and dete
 
 ## Instructions
 
-1. **Read the current mappings** from `VivaDicta/VivaDictaApp.swift` — find the `knownSchemes` dictionary inside `getURLSchemeForBundleId()`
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

PR #381, opened 2026-08-29
Stop seeding the next keyboard session with this session's host

n0an:
## Summary

The keyboard could hand the main app the bundle ID of the app it was in **last** time, teleporting the user into the wrong app. Reproduced manually as a clean off-by-one:

| Step | Persisted value read | Where the user landed |
|---|---|---|
| Telegram, cold start | *(nil)* | nowhere |
| Notes | `ph.telegra.Telegraph` | Telegram |
| Telegram | `com.apple.mobilenotes` | Notes |

`KeyboardContext.hostApplicationBundleId` is persisted to the App Group, so it outlives both the controller and the extension process. Per KeyboardKit's Host article it "will not sync the bundle ID to the `KeyboardContext` unless absolutely necessary" for exactly that reason. We were writing it on every resolve, and nothing in this codebase reads it back (the handoff URL carries the resolved value directly), so the write did nothing but leave this session's host sitting there for the next one to pick up.

Now cleared instead, which also wipes anything an earlier build left behind.

The doc comment asserting that a fresh controller made staleness impossible was wrong on the same point: the persistence lives in the App Group, not the controller.

## Diagnosability

Host resolution and the teleport decision path move from `.info` to `.notice`. Info level is memory backed and dies with the process, and the app is usually terminated moments after a handoff, so a wrong host left no trace in a log collected afterwards. In a recent device capture, four of five handoffs were invisible for this reason.

The resolution log now also reports how long before the resolution the persisted value was last written, which is what distinguishes a stale host from a freshly resolved one.

## Deliberately not included

Rejecting a resolved value whose `hostApplicationBundleIdSyncDate` predates the keyboard session. That would also reject the common case of using the keyboard twice in the same app, and whether the resolver refreshes that date is not verifiable against a binary framework. Shipping it blind risks trading "teleports to the wrong app" for "never teleports."

The logging above is what settles the threshold. The guard is a follow-up gated on one device capture.

## Testing

- `xcodebuild build` against iPhone 17 Pro Max / iOS 26.4: success, 0 errors.
- Behavior verification needs a device capture reproducing the Telegram to Notes to Telegram sequence on a build with these changes, since the failure only shows on a physical device and the relevant logs did not previously persist.

### DIFF
--- a/VivaDicta/VivaDictaApp.swift
+++ b/VivaDicta/VivaDictaApp.swift
@@ -309,11 +309,14 @@ struct VivaDictaApp: App {
         // 2. If we CANNOT return: Start recording → Show return prompt → User manually switches
         //    User sees recording already happening when they arrive
         
-        logger.logInfo("🔄 attemptReturnToHost called with hostId: \(hostId)")
-        logger.logInfo("🔄 Attempting to return to host: \(hostId)")
-        
+        // `.notice` rather than `.info` throughout this path: info-level
+        // entries are memory backed, so they are gone by the time a device log
+        // is collected - and the app is usually terminated moments after a
+        // handoff, which is exactly when a wrong host needs explaining.
+        logger.logNotice("🔄 Attempting to return to host: \(hostId)")
+
         if let url = returnURL(forHostId: hostId) {
-            logger.logInfo("🚀 Found return URL, attempting to open: \(url.absoluteString)")
+            logger.logNotice("🚀 Found return URL, attempting to open: \(url.absoluteString)")
 
             Task {
                 // Check if we have a transcription model selected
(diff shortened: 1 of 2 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Closing in favor of a follow-up. The device capture disproved the diagnosis behind this PR: the persisted App Group value was not the cause. KeyboardKit's resolver returned a bundle ID for an app that had been terminated for three seconds and was never the current host, 1.3s after the keyboard moved to a new app. The real fix is to re-resolve at handoff time rather than reusing the value cached in viewDidLoad. Reopening as a new PR with that fix, the useful hygiene from here, and the sync-age logging dropped (it measured our own write).

PR #382, opened 2026-08-29
Re-resolve the host app at handoff instead of reusing viewDidLoad's answer

n0an:
Replaces #381, whose diagnosis a device capture disproved.

## The bug

Tap the mic in Notes, land in Telegram. The keyboard hands the main app the bundle ID of an app it was not typing into, and the main app dutifully opens it.

## What the capture shows

KeyboardKit's resolver lags a change of host app:

| Time | Event |
|---|---|
| 17:16:21.93 | Telegram process **terminated** |
| 17:16:23.53 | Notes goes **foreground** |
| **17:16:24.88** | keyboard resolves `ph.telegra.Telegraph` |
| 17:16:27.97 | mic tapped, reusing that 3.1s-old value |
| 17:16:28.98 | main app opens `tg://`, **relaunching** Telegram |

The resolved app was neither foreground nor running. Foreground and process-termination times are from SpringBoard and runningboardd in the same archive, so this is system ground truth, not inference.

Every resolution in that capture taken 8s or more after its host appeared was correct. The single wrong one was taken **1.3s** after the host switched.

## The fix

Resolution starts in `viewDidLoad` - the instant the keyboard reaches a new host, and so the least reliable moment to ask. That answer was then cached and reused for a handoff seconds later, by which point a fresh resolve would very likely have been right.

Handoffs now call `hostApplicationBundleIdForHandoff()`, which re-resolves and falls back to the cached answer only if the fresh one cannot be had inside the existing 1s bound. Both handoff call sites already awaited an async accessor, so the shape is unchanged.

## Also included

- **Stop writing `KeyboardContext.hostApplicationBundleId`.** It is persisted to the App Group, KeyboardKit only writes it "unless absolutely necessary" for that reason, and nothing here reads it back, so the write only left this session's host sitting there for the next one. Cleared instead. This was *not* the cause of the bug above, but it is a second route to a stale host and costs nothing to close.
- **Host resolution and the teleport decision path move from `.info` to `.notice`.** Info level is memory backed and dies with the process, and the app is usually terminated moments after a handoff, so a wrong host left no trace in a log collected afterwards. Four of five handoffs in an earlier capture were invisible for this reason.

Dropped from #381: the sync-age suffix on the resolution log. It reported our own `= nil` write, so it measured nothing useful.

## Trade-off

A handoff can now wait up to 1s for a fresh resolve before opening the main app. In practice the resolver returns fast; the bound only bites when it cannot answer at all, which is exactly when the cached value is least trustworthy anyway.

## Testing

- `xcodebuild build`, iPhone 17 Pro Max / iOS 26.4: success, 0 errors.
- Needs a device run of the reproducer: Telegram to Notes to Telegram, force-quitting the app between each, checking you land in the app you started from. The `.notice` logging above makes that verifiable from a collected log for the first time.

Worth reporting upstream against KeyboardKit 10.9.1 separately - a dead, non-foreground bundle ID returned 1.3s after a host switch is a tight repro.

### DIFF
--- a/VivaDicta/VivaDictaApp.swift
+++ b/VivaDicta/VivaDictaApp.swift
@@ -309,11 +309,14 @@ struct VivaDictaApp: App {
         // 2. If we CANNOT return: Start recording → Show return prompt → User manually switches
         //    User sees recording already happening when they arrive
         
-        logger.logInfo("🔄 attemptReturnToHost called with hostId: \(hostId)")
-        logger.logInfo("🔄 Attempting to return to host: \(hostId)")
-        
+        // `.notice` rather than `.info` throughout this path: info-level
+        // entries are memory backed, so they are gone by the time a device log
+        // is collected - and the app is usually terminated moments after a
+        // handoff, which is exactly when a wrong host needs explaining.
+        logger.logNotice("🔄 Attempting to return to host: \(hostId)")
+
         if let url = returnURL(forHostId: hostId) {
-            logger.logInfo("🚀 Found return URL, attempting to open: \(url.absoluteString)")
+            logger.logNotice("🚀 Found return URL, attempting to open: \(url.absoluteString)")
 
             Task {
                 // Check if we have a transcription model selected
@@ -342,14 +345,14 @@ struct VivaDictaApp: App {
 
                 // Now return to the host app
                 if await UIApplication.shared.open(url) {
-                    logger.logInfo("✅ Successfully opened host app: \(hostId) with recording started")
+                    logger.logNotice("✅ Successfully opened host app: \(hostId) with recording started")
                 } else {
                     logger.logError("❌ Failed to open host app: \(hostId)")
                     appState.showKeyboardReturnPrompt = true
                 }
             }
         } else {
-            logger.logInfo("❌ No URL scheme available for host: \(hostId)")
+            logger.logNotice("❌ No URL scheme available for host: \(hostId)")
             // No URL scheme found - start recording and show keyboard return prompt
             // so user can switch back manually and find recording already in progress
             Task {
@@ -563,17 +566,17 @@ struct VivaDictaApp: App {
     /// Returns to the host app without starting recording.
     /// Used by the text processing keyboard flow.
     private func returnToHost(hostId: String) {
-        logger.logInfo("🔄 Returning to host app (no recording): \(hostId)")
+        logger.logNotice("🔄 Returning to host app (no recording): \(hostId)")
 
         guard let url = returnURL(forHostId: hostId) else {
-            logger.logInfo("❌ No return URL for host: \(hostId)")
+            logger.logNotice("❌ No return URL for host: \(hostId)")
             appState.showKeyboardReturnPrompt = true
             return
         }
 
         Task {
             if await UIApplication.shared.open(url) {
-                logger.logInfo("✅ Returned to host app: \(hostId)")
+                logger.logNotice("✅ Returned to host app: \(hostId)")
             } else {
                 logger.logError("❌ Failed to open host app: \(hostId)")
                 appState.showKeyboardReturnPrompt = true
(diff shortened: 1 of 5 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
Pushed `ec19e5813`: dropped the cached-host fallback and widened the resolve bound to 2s.

On a fresh resolve that does not land in time, the handoff now returns nil rather than the `viewDidLoad` answer. That answer is the least trustworthy value the keyboard holds and is wrong in precisely the case the re-resolve exists to handle. Nil degrades safely and was verified end to end: `keyboardHandoff` falls back to the bare `vivadicta://record-for-keyboard`, the app opens, hot mic activates, the no-hostId branch starts recording and shows the manual return prompt. Never a dead button.

The bound goes 1s to 2s since waiting is now the only route to an answer. It was never hit in the device capture, so no added latency in practice.

The `viewDidLoad` resolve stays, but only to activate KeyboardKit's resolver early so the handoff call reads a warm one; its value is no longer used. `hostApplicationBundleId(waitingUpTo:)` is now private.

Device verification of the previous commit, for the record - three taps, one a genuine save:

| Tap | Host | Cached (viewDidLoad) | Fresh (at tap) | Opened |
|---|---|---|---|---|
| 1 | Telegram | `Telegraph` | `Telegraph` | Telegram ✅ |
| 2 | Notes | `mobilenotes` | `mobilenotes` | Notes ✅ |
| 3 | Telegram | `com.apple.Spotlight` | `Telegraph` | Telegram ✅ corrected |

Tap 3 is the morning's failure mode: Telegram opened from Spotlight, `viewDidLoad` latched `com.apple.Spotlight`, and the re-resolve 3.8s later corrected it. Without this change that tap would have failed. No timeout lines in the capture.

n0an:
Fixed in `88317a9a8` - good catch, this was a real bug and it defeated the whole point of the bound.

Confirmed the analysis: `withTaskGroup` awaits every child before returning, and `group.cancelAll()` cancels the child but not the unstructured `hostApplicationBundleIdTask` it is awaiting. `Task<String?, Never>.value` has nothing to throw on cancellation, so the losing child kept waiting for the resolver regardless. With every handoff now starting a fresh resolution, a slow resolver would have held both the mic and text-processing deep links open indefinitely - an apparently inert tap, which is worse than the stale host this PR set out to fix.

Rather than repair the race, I removed it. KeyboardKit's `resolveHostApplicationBundleId(interval:timeout:)` takes a timeout and, per its docs, "will poll for a result until the timeout is reached" - so the deadline now lives with the resolver and the task is awaited directly. `hostApplicationBundleId(waitingUpTo:)` and the task group are deleted.

Residual risk worth naming: the bound is now only as good as KeyboardKit's honoring of its own `timeout:`. If it ignores it, the handoff can still stall. That is a smaller and more visible surface than the previous code, which could not bound anything even with a well-behaved resolver, and the failure is now upstream's rather than ours.

Build clean. The device verification in the comment above was against `c70a942`; this changes the bounding mechanism but not the re-resolve behavior that verification exercised.

n0an:
Thanks - reviewed against current HEAD, since this review was written against `c70a942` and two commits have landed since.

**1. Error path skipped the clear** - valid, fixed in `b21fbfde6`. Moved the clear ahead of the resolve so it runs unconditionally rather than being skipped by an early `return nil` in the `catch`. You were right that it silently contradicted the comment directly above it.

**2. "Timed out" message fires on a genuine nil** - already resolved by `ec19e5813` and `88317a9a8`. There is no cached fallback any more, so a fresh `nil` is simply returned rather than being overridden by a stale value, and the message is now `🏠 Host unresolved; handing off without one`, which no longer asserts a timeout that may not have happened. Both halves of the concern are gone.

Worth noting one of your premises is now outdated in a way that mattered: the review says "latency stays bounded at ~1s". Codex found that it was not bounded at all - a task group awaits every child before returning, and cancelling the losing child does not cancel the unstructured Task it awaits. That is fixed in `88317a9a8` by handing the deadline to KeyboardKit's `resolveHostApplicationBundleId(timeout:)` and deleting the race.

**Minor: stale docs** - confirmed, `Deep-Linking-URL-Routing-Architecture.md:80` and `Keyboard-Extension-Architecture.md:303-308` still describe `hostApplicationBundleId` as a synchronous property and quote the old inline call site. Leaving that decision to the author rather than expanding this PR's scope unasked.

PR #385, opened 2026-09-06
Auto-enable a newly added keyboard language on update

n0an:
## Summary

The keyboard seeds its enabled-languages set from the user's iOS language list exactly once, inside `ensureLanguageSettingsMigrated()`, behind the `didMigrateLanguageSettings_v1` boolean. Every install that has launched 3.2.x or later already has that flag set, so the seeding never runs again - a language shipped in a later release stays off no matter what the user has configured in iOS Settings.

Czech (#383) landed straight into that hole. A Czech user updating the app would have had to find the toggle by hand, and the same would have been true of every language added after it. `case .czech: ["cs"]` was correct and completely inert for the existing user base.

## What changed

Replace the one-shot boolean with a record of which languages have already been *offered*. `ensureNewLanguagesSeeded()` diffs `allCases` against that record on every language read and considers only the difference:

- A new case is auto-enabled when it matches the user's system languages.
- A language the user deliberately switched off is never switched back on - it is already in the offered record.
- Each language is evaluated exactly once, whether or not it ended up enabled, so adding Czech to iOS Settings *after* the update does not resurrect a language the user already rejected.
- Seeding only ever adds, so the English floor is untouched.

Installs that migrated under an older build have no such record, so they backfill from `preSeedingLanguages` - the set that shipped before this key existed (`english, french, german, spanish, russian`). That constant is a frozen historical fact rather than a list to maintain: a new language is picked up precisely *by being absent from it*, and the doc comment says so.

The system list is read through an `@autoclosure`, so `Locale.preferredLanguages` is not touched on the common path where there is nothing new to seed. This runs on every language read, including from the keyboard extension.

Two pieces of cleanup along the way:
- The comma-separated parse/serialize of the language set was written out three times (getter, setter, migration). Folded into `parseLanguages` / `serializeLanguages`.
- `preferredFromSystem()` now takes an injectable tag list (defaulting to the live one) so the locale matching is testable.

## Testing

New `KeyboardLanguageSeedingTests` (12 tests) covering the bug and its edges: a new language enabled for an already-migrated install, skipped when it does not match the system list, an already-offered language never re-enabled, each language considered only once across two passes, seeding staying additive, the bookkeeping key, the migration marking everything offered on a fresh install, canonical serialization order, unknown raw values dropped, and region-subtag/unsupported-language matching.

- `swift test` in `Modules/AppGroup`: 37 tests, all passing.
- `xcodebuild build` for the app scheme: success, 0 errors.

## Notes

No schema or CloudKit impact - this is UserDefaults in the app group. Behavior for a fresh install is unchanged: the migration seeds from the system list and marks everything offered, so seeding finds nothing to do.

### DIFF
--- a/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
+++ b/Modules/AppGroup/Sources/AppGroup/AppGroupCoordinator.swift
@@ -70,6 +70,10 @@ public final class AppGroupCoordinator: @unchecked Sendable {
     public static let kEnabledKeyboardLanguages = "enabledKeyboardLanguages"
     public static let kCurrentKeyboardLanguage = "currentKeyboardLanguage"
     private static let kDidMigrateLanguageSettings = "didMigrateLanguageSettings_v1"
+    /// Languages already offered to this install, so a language added in a
+    /// later release can be auto-enabled exactly once. See
+    /// `ensureNewLanguagesSeeded()`.
+    private static let kSeededLanguages = "seededKeyboardLanguages"
     // Legacy keys, kept here only so the migration code can read and clear them.
     private static let kLegacyKeyboardLayoutStyle = "keyboardLayoutStyle"
     private static let kLegacyIsRussianLayoutEnabled = "isRussianLayoutEnabled"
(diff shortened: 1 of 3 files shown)


n0an:
@claude please review this PR

n0an:
@codex please review this PR

n0an:
@claude please review this PR

